PR #25, opened 2019-05-05
Add support for building with Meson

randy408:


src/world.c
```diff
@@ -1236,7 +1236,6 @@ ecs_entity_t ecs_import_from_library(
 #else
     ecs_os_err(
         "sorry, loading libraries is only possible if flecs is built with bake :(");
-    ut_raise();
```
ArnCarveris: @SanderMertens It's okay to remove maybe there should be `ecs_os_abort`, or new os API function `ecs_os_raise` ?

SanderMertens: ut_raise() is specific to the logging framework in `bake.util`. I'd prefer to keep the flecs logging callbacks as simple as possible.

This call can be removed.

PR #29, opened 2019-05-18
rename include/util to include/flecs

randy408:
This change renames include/util to be more flecs-specific to avoid naming conflicts when the directory is installed to e.g. `/usr/include`.


randy408:
This is now outdated.

Issue #90, opened 2019-10-23
Use more descriptive error messages when signature parsing fails


SanderMertens:
**Describe the problem you are trying to solve.**
When parsing of a system signature fails, Flecs will return an "invalid system signature" message, often not specifying what is not valid about the signature. This makes debugging such errors much more difficult.

**Describe the solution you'd like**
More descriptive error messages for the signature (and type) parser.

WEREMSOFT:
Can you give a step-by-step way to reproduce an error to start with?  Ideally a piece of code. Thanks

SanderMertens:
Sure thing. Right now, all error handling in Flecs is done with error codes (https://github.com/SanderMertens/flecs/blob/master/include/flecs.h#L2122) that are passed to either `ecs_abort` or `ecs_assert`. This guarantees consistent error messages, but the API has as downside that there is not much room for additional context.

This is especially annoying with the parser of system signatures. Consider the following code example:

```c
void Move(ecs_rows_t *rows) { }

int main(int argc, char *argv[]) {
    ecs_world_t *world = ecs_init_w_args(argc, argv);
    ECS_SYSTEM(world, Move, EcsOnUpdate, Position, Velocity);
}
```

This yields the following error:
```
[err] abort .../flecs/src/system.c:437: invalid component id (Position)
```

While it correctly reports that `Position` is not defined, it does not tell me for which system. When you have lots of systems, this can be tricky to debug.

In some cases, like this one, it will throw a generic "invalid expression" error:

```c
typedef struct Position {float x, y; } Position;

void Move(ecs_rows_t *rows) { }

int main(int argc, char *argv[]) {
    ecs_world_t *world = ecs_init_w_args(argc, argv);
    ECS_COMPONENT(world, Position);
    ECS_SYSTEM(world, Move, EcsOnUpdate, .);
}
```

It would be nice if these errors would be extended with at least the system name and the position in the signature at which the error occurred. I think it would be best if this functionality would be built specifically for the parser, to prevent bloating the existing error APIs.

WEREMSOFT:
I'm going to start to work in this issue today. Can we flag it as "taken" or "in process"? Thanks.

WEREMSOFT:
I created a PR. Please take a look:

https://github.com/SanderMertens/flecs/pull/99

WEREMSOFT:
I updated the PR. Can you please take a look before I update the tests? 

https://github.com/SanderMertens/flecs/pull/99

Thanks.

[...]

PR #99, opened 2019-11-04
#90, adding more descriptive error message when adding a system.

WEREMSOFT:
ANSI C is not one of my strengths. 

I tried to not modify the interfaces but it was hard to send the system id through the the call chain.

I'm not sure about the solution. It's already an improvement, but I don't like to polute the API with parameters that are only used to display an error. So I only add the changes in parser.c.

There are chances to modify the function "ecs_parse_signature_action" in order to get the system_id and process the error. But if I do this. I have to modify also the function "has_tables", which does not report any error. So, the system_id will be an unused variable.

Can you please take a look before I update the tests? This are the use cases:

### Use Case 1 
```
#include <stdio.h>
#include "flecs.h"

typedef struct Position {float x, y; } Position;
typedef struct Velocity {float x, y; } Velocity;

void Move(ecs_rows_t *rows) { }

int main(int argc, char *argv[]) {
    ecs_world_t *world = ecs_init_w_args(argc, argv);
    ECS_COMPONENT(world, Position);
    ECS_SYSTEM(world, Move, EcsOnUpdate, Position, Velocity);
}
```
## Output:
```
Move at argument #2. Error: "unresolved component name 'Velocity'"
Position, Velocity
~~~~~~~~~~^
```

### Use Case 2
```
#include <stdio.h>
#include "flecs.h"

typedef struct Position {float x, y; } Position;
typedef struct Velocity {float x, y; } Velocity;

void Move(ecs_rows_t *rows) { }

int main(int argc, char *argv[]) {
    ecs_world_t *world = ecs_init_w_args(argc, argv);
    ECS_COMPONENT(world, Position);
    ECS_SYSTEM(world, Move, EcsOnUpdate, Position | !Velocity);
}
```
## Output:
```
Move at argument #1. Error: "cannot use NOT (!) operator in an OR (|) expression"
Position | !Velocity
~~~~~~~~~~~~^
flecs/src/parser.c:239: invalid type expression/signature
```
### Use Case 3
```
#include <stdio.h>
#include "flecs.h"

typedef struct Position {float x, y; } Position;
typedef struct Velocity {float x, y; } Velocity;

void Move(ecs_rows_t *rows) { }

int main(int argc, char *argv[]) {
    ecs_world_t *world = ecs_init_w_args(argc, argv);
    ECS_COMPONENT(world, Velocity);
    ECS_SYSTEM(world, Move, EcsOnUpdate, Velocity, .);
}
```
## Output:
```
flecs/src/parser.c:69: invalid type expression/signature (Velocity, .)
Move at argument #2. Error: "invalid type expression/signature"
Velocity, .
~~~~~~~~~~^
```
### Use Case 4
```
#include <stdio.h>
#include "flecs.h"

typedef struct Position {float x, y; } Position;
typedef struct Velocity {float x, y; } Velocity;

void Move(ecs_rows_t *rows) { }

int main(int argc, char *argv[]) {
    ecs_world_t *world = ecs_init_w_args(argc, argv);
    ECS_COMPONENT(world, Velocity);
    ECS_SYSTEM(world, Move, EcsOnUpdate, .);
}
```
## Output:
```
flecs/src/parser.c:69: invalid type expression/signature (.)
Move at argument #1. Error: "invalid type expression/signature"
.
^
```

### Known bugs
Sometimes, the error and the "printf" appear mixed. I'm not sure how to "flush" the printf output before printing the error.


src/column_system.c
```diff
@@ -763,7 +763,7 @@ ecs_entity_t ecs_new_col_system(
     system_data->base.kind = kind;
     system_data->base.cascade_by = 0;
     system_data->base.has_refs = false;
-    system_data->base.needs_tables = ecs_needs_tables(world, sig);
+    system_data->base.needs_tables = ecs_needs_tables(world, sig, id);
```
SanderMertens: Suggestion: pass the system entity id (`result`) to the function. That way you can access all the information you need. You can get the id from a system like this:

```c
const char* id = ecs_get(world, result, EcsId);
```

src/column_system.c
```diff
@@ -776,8 +776,10 @@ ecs_entity_t ecs_new_col_system(
     system_data->inactive_tables = ecs_vector_new(
         &matched_table_params, ECS_SYSTEM_INITIAL_TABLE_COUNT);
 
+    printf("System name %s\n", *id_data);
+    // TODO: This funciton should return -1 and the component error should be handled here.
     ecs_parse_component_expr(
-        world, sig, ecs_parse_signature_action, system_data);
+        world, sig, ecs_parse_signature_action, system_data, id);
```
SanderMertens: Same here: I'd pass in the `result`.

src/parser.c
```diff
@@ -101,10 +101,11 @@ int has_tables(
 /* Does expression require that a system matches with tables */
 bool ecs_needs_tables(
     ecs_world_t *world,
-    const char *signature)
+    const char *signature,
+    const char *systen_id)
```
SanderMertens: Typo: `system_id`

src/flecs_private.h
```diff
@@ -425,17 +425,22 @@ ecs_row_t ecs_to_row(
...
+void build_error_string(char *error, const char* signature, long position, const char *system_id);
+
 /* Utility that parses system signature */
 int ecs_parse_component_expr(
     ecs_world_t *world,
     const char *sig,
     ecs_parse_action_t action,
-    void *ctx);
+    void *ctx,
+    const char *system_id);
```
SanderMertens: This is a bit nitpicking- but I usually keep a `ctx` argument as last argument of a function. I'd change it to the system entity handle, and move it to one spot before `ctx`.

src/flecs_private.h
```diff
@@ -425,17 +425,22 @@ ecs_row_t ecs_to_row(
...
 
 /* Test whether signature has columns that must be retrieved from a table */
 bool ecs_needs_tables(
     ecs_world_t *world,
-    const char *signature);
+    const char *signature,
+    const char *system_id);
```
SanderMertens: Typo: `systen_id`

src/parser.c
```diff
@@ -170,13 +171,18 @@ const char* parse_annotation(
 
     return ptr;
 }
+/** Build error string*/
+void build_error_string(char *error, const char* signature, long position, const char *system_id){
+    sprintf(error, "\"%s\" at position %ld. System \"%s\"", signature, position, system_id);
```
SanderMertens: This looks good, already much more informative! What I'd suggest is to add a descriptive string and format it so that it looks a bit more parser-ish, something like this:

```
Move: 1: no such component "Position"
    Position, Velocity
    ^
```

src/parser.c
```diff
@@ -196,7 +202,10 @@ int ecs_parse_component_expr(
 
         if (prev_is_0) {
             /* 0 can only apppear by itself */
-            ecs_abort(ECS_INVALID_SIGNATURE, sig);
+            char err[50] = "/0";
+            build_error_string(err, sig, ptr - sig, system_id);
```
SanderMertens: Example description: "0 can only appear by itself"

src/parser.c
```diff
@@ -225,25 +234,33 @@ int ecs_parse_component_expr(
                 ecs_system_expr_oper_kind_t prev_oper_kind = oper_kind;
                 bptr = parse_complex_elem(bptr, &elem_kind, &oper_kind, &source);
                 if (!bptr) {
-                    ecs_abort(ECS_INVALID_EXPRESSION, sig);
+                    char err[50] = "/0";
```
SanderMertens: If this happens, an error was detected in the `parse_complex_elem` function. The error should be printed there since we have more information there about what went wrong.

src/parser.c
```diff
@@ -196,7 +202,10 @@ int ecs_parse_component_expr(
 
         if (prev_is_0) {
             /* 0 can only apppear by itself */
-            ecs_abort(ECS_INVALID_SIGNATURE, sig);
+            char err[50] = "/0";
+            build_error_string(err, sig, ptr - sig, system_id);
+
+            ecs_abort(ECS_INVALID_SIGNATURE, err);
```
SanderMertens: I think it's nicer to first print the error (in a format like I described above) and then call the abort. Otherwise the error message gets appended to the abort, and you get very long error messages which likely won't fit on a single line.

So something like:

```c
print_error_string(system_id, sig, "Unknown component '%s'", component_id);
ecs_abort(ECS_INVALID_SIGNATURE, NULL);
```

src/parser.c
```diff
@@ -225,25 +234,33 @@ int ecs_parse_component_expr(
...
-                    ecs_abort(ECS_INVALID_EXPRESSION, sig);
+                    char err[50] = "/0";
+                    build_error_string(err, sig, ptr - sig, system_id);
+                    ecs_abort(ECS_INVALID_EXPRESSION, err);
                 }
 
                 if (oper_kind == EcsOperNot && prev_oper_kind == EcsOperOr) {
-                    ecs_abort(ECS_INVALID_EXPRESSION, sig);
+                    char err[50] = "/0";
+                    build_error_string(err, sig, ptr - sig, system_id);
```
SanderMertens: Suggested description:

"cannot use NOT (!) operator in an OR (|) expression"

src/parser.c
```diff
@@ -225,25 +234,33 @@ int ecs_parse_component_expr(
...
+                    ecs_abort(ECS_INVALID_EXPRESSION, err);
                 }
             }
 
            if (oper_kind == EcsOperOr) {
                 if (elem_kind == EcsFromEmpty) {
                     /* Cannot OR handles */
-                    ecs_abort(ECS_INVALID_EXPRESSION, sig);
+                    char err[50] = "/0";
```
SanderMertens: Suggested description:

"cannot use OR (|) operator in combination with an empty FROM (.) expression"

src/parser.c
```diff
@@ -225,25 +234,33 @@ int ecs_parse_component_expr(
...
+                    build_error_string(err, sig, ptr - sig, system_id);
+                    ecs_abort(ECS_INVALID_EXPRESSION, err);
                 }
             }            
 
             if (!strcmp(bptr, "0")) {
                 if (bptr != buffer) {
                     /* 0 can only appear by itself */
-                    ecs_abort(ECS_INVALID_EXPRESSION, sig);
+                    char err[50] = "/0";
```
SanderMertens: Suggested description:

"0 can only appear by itself"

SanderMertens:
Thanks for the PR! I added some comments to the review. These are just suggestions, let me know what you think!

WEREMSOFT:
Sorry for the late response, I answered your comment but github was waiting for confirmation about my response. Please check my answer about src/column_system.c on line 766. I have problems to implement your suggestion if the system does not already exists in the world.

WEREMSOFT:
By the way, I'm aware of the failed builds. I'll fix them on the next commit.

src/parser.c
WEREMSOFT: Sure. I need to think how to format.

src/flecs_private.h
WEREMSOFT: done.

src/flecs_private.h
WEREMSOFT: Done

src/column_system.c
SanderMertens: Oh, no you're not. Good point, looks like passing the system id is the right thing to do after all.

src/parser.c
WEREMSOFT: done

src/type.c
```diff
@@ -856,7 +856,7 @@ EcsTypeComponent type_from_expr(
     const char *expr)
 {
     ecs_vector_t *vec = ecs_vector_new(&handle_arr_params, 1);
-    ecs_parse_component_expr(world, expr, parse_type_action, &vec);
+    ecs_parse_component_expr(world, expr, parse_type_action, NULL, &vec);
```
WEREMSOFT: I don't like this (NULL in the system_id). What do you think? Can you think in a better way to do it??

src/parser.c
WEREMSOFT: Done.

src/parser.c
WEREMSOFT: done

src/parser.c
WEREMSOFT: done

src/parser.c
WEREMSOFT: done

src/parser.c
WEREMSOFT: Done. Still, the error and the printf are mixed together. Like the error starts to print the message while the ptintf is still working.

src/parser.c
WEREMSOFT: I tried to do it, but It seems to me that I have to pollute the system.c API in order to give the information to communicate que error. Maybe there is a way, but I'd prefer to leave it for later, once I have a better mental image of the project. For now I just tried to touch only the parser. If you have a idea about how to handle errors at system level, I'm open for suggestions.

src/type.c
SanderMertens: I think for now it's ok. Perhaps in a later revision `system_id` is not the right name, and it should be something more generic like `name`.

src/parser.c
```diff
@@ -203,10 +203,8 @@ int ecs_parse_component_expr(
 
         if (prev_is_0) {
             /* 0 can only apppear by itself */
-            char err[50] = "0 can only apppear by itself";
-            ecs_print_error_string(err, sig, system_id, bptr);
-
-            ecs_abort(ECS_INVALID_SIGNATURE, err);
+            ecs_print_error_string(ecs_strerror(ECS_ZERO_CAN_ONLY_APPEAR_BY_ITSELF), sig, system_id, bptr);
```
SanderMertens: I think (but I can be persuaded if there are good reasons not to) that I would prefer a printf-style formatted string here, so that you could do something like (example):

```
ecs_print_error_string(sig, system_id, bptr, "unresolved component name '%s'", component_name);
```
In the current setup you print the same error twice (in the `ecs_print_error_string` and in the `ecs_abort`) and the error message is constrained by what limited parameters you can pass into either.

WEREMSOFT: I need  to turn `ecs_print_error_string` into a variadic macro. 

I'll try to send a PR tonight.

WEREMSOFT: Done, please take a look. Now the message is 
```
Move at argument #2. Error: "unresolved component name 'Velocity'"
Position, Velocity
~~~~~~~~~~^
[err] abort /Users/pabloweremczuk/Documents/Proyectos/cpp/flecs/src/parser.c:276: invalid component id (Velocity)
```

SanderMertens: That looks awesome! taking a look

WEREMSOFT: Please let me know if more changes are required. My plan is to work on the tests this weekend.

Thanks.

include/flecs.h
```diff
@@ -2362,7 +2367,43 @@ void _ecs_assert(
 
 /** Calculate offset from address */
 #define ECS_OFFSET(o, offset) (void*)(((uintptr_t)(o)) + ((uintptr_t)(offset)))
+#define is_empty(...) ( sizeof( (char[]){#__VA_ARGS__} ) == 1 )
```
SanderMertens: Is this portable?

include/flecs.h
```diff
@@ -2362,7 +2367,43 @@ void _ecs_assert(
...
 
+const char* if_encapsulation(const char * component_id, const char* signature, uint *position) {
```
SanderMertens: You should not add functions in header files.

include/flecs.h
```diff
@@ -2362,7 +2367,43 @@ void _ecs_assert(
...
+            char* pointer_to_string = strstr(signature, component_id);
+            *position = pointer_to_string - signature;
+        }
+        return component_id;
+    }
+
+    return signature;
+}
+
+#define ecs_print_error_string(signature, system_id, error_description, component_id, ...)\
```
SanderMertens: No need to make this a macro, you can change thxs to a function with a variable argument list:

```c
void ecs_print_error_string(signature, system_id, error_description, component_id, ...)
```

include/flecs.h
WEREMSOFT: I remove it since I'm not using macros anymore.

include/flecs.h
WEREMSOFT: removed

include/flecs.h
WEREMSOFT: removed

Issue #70, opened 2019-09-09
compiler warnings with Visual Studio 2017


mattiasljungstrom:
I've been evaluating this library as one possible solution for a project, but I noticed there are quite a lot of warnings when compiling with visual studio 2017. 

Examples:
flecs\src\column_system.c(281): warning C4146: unary minus operator applied to unsigned type, result still unsigned
flecs\src\column_system.c(934): warning C4244: '=': conversion from 'uint64_t' to 'uint32_t', possible loss of data
flecs\src\column_system.c(1001): warning C4244: '+=': conversion from 'double' to 'float', possible loss of data
flecs\src\entity.c(1050): warning C4018: '<': signed/unsigned mismatch
flecs\src\entity.c(300): warning C4090: 'function': different 'const' qualifiers
flecs\src\type.c(1068): warning C4244: 'initializing': conversion from 'ecs_entity_t' to 'int', possible loss of data
etc...

Would you be interested in a push request if I keep using this library?

SanderMertens:
@mattiasljungstrom Sure thing! I just built Flecs with pedantic warnings on MacOS and fixed a few warnings that showed up there, but it seems like msvc is catching more. If you can do a PR I'll definitely merge it.

SanderMertens:
I've recently fixed a lot of warnings in MSVC related to type conversions and "unsafe" functions, so closing this issue. If you are still seeing issues with MSVC, feel free to reopen.

PR #107, opened 2019-11-08
#70 fix VS 2019 compiler warnings

shrub77:
I've removed compiler warnings in MSVC using VS 2019. Most of the warnings are related to implicit type conversions, which are now explicit, though I suggest they be scrutinized as I wasn't sure in some cases. Added pragmas to ignore certain warnings related to alloca usage and static analysis in malloc'd arrays


.gitignore
```diff
@@ -3,3 +3,143 @@ bin
 .bake_cache
 .vscode
 gcov
+/build/ZERO_CHECK.vcxproj.filters
```
SanderMertens: I think this was accidentally checked in? If you'd like to ignore everything in the build directory, you can just add "build" to the `.gitignore` file.

src/entity.c
```diff
@@ -412,12 +414,12 @@ ecs_type_t copy_from_prefab(
         }
 
         ecs_table_column_t *src_column = &prefab_columns[p + 1];
-        uint32_t size = src_column->size;
+        uint64_t size = src_column->size;
```
SanderMertens: This would a bit excessive for a column (component type) size. The size in the type itself is `uint16_t`, I would use the same here.

src/entity.c
```diff
@@ -412,12 +414,12 @@ ecs_type_t copy_from_prefab(
...
 
         if (size) {
             void *src_column_data = ecs_vector_first(src_column->data);
             void *src_ptr = ECS_OFFSET(
-                src_column_data, size * (prefab_index - 1));
+                src_column_data, size * ((uint64_t)prefab_index - 1));
```
SanderMertens: After updating the size variable, hopefully the need for this cast will go away.

src/entity.c
```diff
@@ -777,17 +779,19 @@ void* get_ptr_from_prefab(
         if (prefab == previous) {
             continue;
         }
+		ecs_entity_info_t prefab_info = { .entity = prefab };
```
SanderMertens: Indentation

src/entity.c
```diff
@@ -2278,8 +2282,7 @@ ecs_entity_t ecs_new_component(
     }
 
     result = _ecs_new(world, world->t_component);
-    ecs_set(world, result, EcsComponent, {.size = size});
+    ecs_set(world, result, EcsComponent, {.size = (const uint32_t)size});
```
SanderMertens: Can you explain the need for a `const uint32_t` cast?

src/map.c
```diff
@@ -232,7 +232,7 @@ void resize_map(
     uint32_t bucket_count)
 {
     uint32_t *old_buckets = map->buckets;
-    uint32_t old_bucket_count = map->bucket_count;
+    uint32_t old_bucket_count = (uint32_t)map->bucket_count;
```
SanderMertens: I would make the `old_bucket_count` variable of type `size_t` so that you don't need a cast.

src/map.c
```diff
@@ -311,10 +311,10 @@ void* _ecs_map_set(
     (void)size;
     ecs_assert(ecs_map_data_size(map) == size, ECS_INVALID_PARAMETER, NULL);
 
-    uint32_t bucket_count = map->bucket_count;
+    uint32_t bucket_count = (uint32_t)(map->bucket_count);
```
SanderMertens: Same as above

src/map.c
```diff
@@ -441,15 +441,15 @@ uint32_t ecs_map_count(
 uint32_t ecs_map_bucket_count(
```
SanderMertens: I would change the return type of this function to `size_t`

src/map.c
```diff
@@ -574,5 +574,5 @@ void* ecs_map_next_w_size(
 uint32_t ecs_map_data_size(
```
SanderMertens: I would change the return type of this function to `size_t`

src/misc.c
```diff
@@ -30,8 +30,8 @@ void ecs_sleepf(
     double t)
 {
     if (t > 0) {
-        int sec = t;
-        int nsec = ((t - sec) * 1000000000);
+        int sec = (int)t;
```
SanderMertens: Are implicit conversions not allowed in MSVC?

src/system.c
```diff
@@ -259,8 +259,10 @@ void register_out_columns(
         if (columns[i].inout_kind == EcsOut) {
             if (!system_data->on_demand) {
                 system_data->on_demand = ecs_os_malloc(sizeof(ecs_on_demand_out_t));
-                system_data->on_demand->system = system;
-                system_data->on_demand->count = 0;
+				if (system_data->on_demand != NULL) {
```
SanderMertens: Indentation. Also, while it is good to check for the return of malloc, the pattern in the code has been to assert if it returns `NULL` with a
```c
ecs_assert(system_data->on_demand != NULL, ECS_OUT_OF_MEMORY, NULL);
```

The reasoning behind this is that otherwise the code is left in an undefined state (the system has not been properly created) from which it cannot recover.

src/table.c
```diff
@@ -31,7 +31,7 @@ ecs_table_column_t* new_columns(
     ecs_type_t type)
 {
     ecs_table_column_t *result = ecs_os_calloc(
-        sizeof(ecs_table_column_t), ecs_vector_count(type) + 1);
+        sizeof(ecs_table_column_t), (uint64_t)ecs_vector_count(type) + 1);
```
SanderMertens: Why is a cast to `uint64_t` needed here? The argument is of type `size_t`.

src/type.c
```diff
@@ -533,7 +535,7 @@ ecs_type_t ecs_type_add_intern(
     ecs_assert(world != NULL, ECS_INTERNAL_ERROR, NULL);
 
     uint32_t count = ecs_vector_count(type);
-    ecs_entity_t *new_array = ecs_os_alloca(ecs_entity_t, count + 1);
+    ecs_entity_t *new_array = ecs_os_alloca(ecs_entity_t, (uint64_t)count + 1);
```
SanderMertens: Same as above

src/type.c
```diff
@@ -613,7 +615,7 @@ ecs_type_t ecs_type_merge_intern(
     ecs_entity_t *buf_new = NULL; 
     if (cur_count + add_count) {
         ecs_assert((cur_count + add_count) < ECS_MAX_ENTITIES_IN_TYPE, ECS_TYPE_TOO_LARGE, NULL);
-        buf_new = ecs_os_alloca(ecs_entity_t, cur_count + add_count);
+        buf_new = ecs_os_alloca(ecs_entity_t, (size_t)cur_count + (size_t)add_count);
```
SanderMertens: Yay, here you cast to `size_t`!

src/type.c
```diff
@@ -1067,11 +1069,11 @@ char* ecs_type_to_expr(
         if (id) {
             str = *id;
         } else {
-            int h_int = h;
+            int h_int = (int)h;
             sprintf(buf, "%u", h_int);
             str = buf;
         }
-        len = strlen(str);
+        len = (uint32_t)(strlen(str));
```
SanderMertens: Suggest to change the type of `len` to `size_t`

src/world.c
```diff
@@ -1584,5 +1584,5 @@ uint32_t ecs_get_threads(
 uint32_t ecs_get_target_fps(
```
SanderMertens: The return type of this function should be changed to `float`.

SanderMertens:
Thanks for the changes! I left some comments. As a general thought I would like to prevent excessive casting if it can be avoided by changing the type of a variable. The code sometimes uses conflicting types (thanks for pointing that out) and it is better to fix one of the conflicts than to add lots of casts.

This will keep the code more maintainable, as the compiler will throw warnings when issues with types occur, whereas with a cast it is very easy to mask the real problem

src/entity.c
shrub77: The issue is that ECS_OFFSET casts to uintptr_t which is 8 bytes, so it gives this warning:


Warning	C26451	Arithmetic overflow: Using operator '*' on a 4 byte value and then casting the result to a 8 byte value. Cast the value to the wider type before calling operator '*' to avoid overflow (io.2).

src/entity.c
shrub77: Should just be a uint32_t.

src/misc.c
shrub77: The implicit cast converts double to an int, warns about loss of data:

Warning	C4244	'initializing': conversion from 'double' to 'int', possible loss of data	

imo, this sort of casting should always be explicit

SanderMertens:
The issues fixed in this PR mostly address signed/unsigned mismatches, and the way they are solved is through a cast. The actual problem was that the types were used inconsistently in the framework. For the next minor release branch (bleeding_edge) I addressed these issues by changing (almost) all integer types to signed, which is a bit cleaner as it doesn't require casting.

Issue #115, opened 2019-12-16
Implement modern C++ API


SanderMertens:
**Describe the problem you are trying to solve.**
The current C API is usable from C++, but it does not follow modern C++ practices. Furthermore, some features can be implemented in a cleaner/more type safe way using modern C++ features.

**Describe the solution you'd like**
A type safe, modern C++ wrapper (at least C++11) around the C API.

PR #118, opened 2019-12-29
#115 C++ api

SanderMertens:


include/flecs/flecs.hpp
```diff
@@ -0,0 +1,1571 @@
...
+        ecs_assert(index < m_count, ECS_COLUMN_INDEX_OUT_OF_RANGE, NULL);
+        return m_array[index];
+    }
+
+    T* operator->() {
+        ecs_assert(m_array != nullptr, ECS_COLUMN_INDEX_OUT_OF_RANGE, NULL);
+        return m_array;
+    }
+
+    bool is_set() {
```
Milerius: this function could be marked as cont

include/flecs/flecs.hpp
```diff
@@ -0,0 +1,1571 @@
...
+static const entity_t Singleton = EcsSingleton;
+static const entity_t Invalid = EcsInvalid;
+
+
+////////////////////////////////////////////////////////////////////////////////
+//// Wrapper class around a table column
+////////////////////////////////////////////////////////////////////////////////
+
+template <typename T>
+class column final {
```
Milerius: Why the column is marked as final, without inheritance ?

include/flecs/flecs.hpp
```diff
@@ -0,0 +1,1571 @@
...
+public:
+    column(T* array, size_t count, bool is_shared = false)
```
Milerius: You should prefer using `std::size_t` for c++ version of C type, (<cstddef>)

include/flecs/flecs.hpp
```diff
@@ -0,0 +1,1571 @@
...
+    T* operator->() {
+        ecs_assert(m_array != nullptr, ECS_COLUMN_INDEX_OUT_OF_RANGE, NULL);
+        return m_array;
+    }
+
+    bool is_set() {
+        return m_array != nullptr;
+    }
+
+    bool is_shared() {
```
Milerius: Can also be const

include/flecs/flecs.hpp
```diff
@@ -0,0 +1,1571 @@
...
+    type table_type() const;
+
+    /* Obtain untyped pointer to table column */
+    void* table_column(uint32_t table_column) {
+        return ecs_table_column(m_rows, table_column);
+    }
+
+    /* Obtain typed pointer to table column */
+    template <typename T>
+    flecs::column<T> table_column() {
```
Milerius: I think this function can be marked as const

include/flecs/flecs.hpp
SanderMertens: I prevent inheriting from framework classes wherever possible, just because it's bad practice.

PR #142, opened 2020-01-18
Add test for OnAdd system with entity reference

ArnCarveris:


SanderMertens:
This PR is no longer valid as a result of the upcoming changes in v2

PR #178, opened 2020-04-11
Small C++ Perf Opt: Mark invoker calls to capture by reference 

ikrima:
and mark func args as T&&


SanderMertens:
Awesome, thanks!

PR #193, opened 2020-07-07
Fix static builds with Meson

randy408:


SanderMertens:
Thanks!

randy408:
I forgot about `flecs_EXPORTS` the first time, it's now tested to work with mingw and msvc (shared/static).

PR #246, opened 2020-09-14
V2 minor fixes

ikrima:
Porting several minor fixes from bleeding edge and combined them in a single PR:

- _*(needs checking)*_ ecs_merge: should assert if world->in_progress == false
- bug: flecs::world move constructor doesn't "steal" from passed in world causing double destruction
- fix: mark single parameter constructors as explicit to prevent incorrect automatic type conversion
- ~~lint: id_no_lifecycle() - default nullptr value for world. Function asserts if its nullptr ergo that's can never be a valid value~~
- fix: entity::set() should not take a const rvalue parameter as that prevents moves

### DIFF
--- a/include/flecs/flecs.hpp
+++ b/include/flecs/flecs.hpp
@@ -1516,7 +1516,7 @@ public:
      * @param value The value to assign to the component.
      */
     template <typename T>
-    const base_type& set(const T&& value) const {
+    const base_type& set(T&& value) const {
         static_cast<base_type*>(this)->invoke(
         [&value](world_t *world, entity_t id) {
             ecs_set_ptr_w_entity(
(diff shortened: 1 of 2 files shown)


include/flecs/flecs.hpp
```diff
@@ -2855,7 +2863,7 @@ class component_info final {
         s_name = ecs_get_fullpath(world, entity);
     }
 
-    static entity_t id_no_lifecycle(world_t *world = nullptr, const char *name = nullptr) {
+    static entity_t id_no_lifecycle(world_t *world, const char *name = nullptr) {
```
SanderMertens, line 2866: The `world = nullptr` shouldn't be removed. It is possible for an application to request an id without providing the world (for example, using `flecs::type_id<T>()`).

ikrima: Reverted

PR #247, opened 2020-09-15
V2 minor fixes II

ikrima:
- ecs_os_api_strdup() default implementation incorrectly using default ecs_os_api_malloc
    should go through override path
- fix: don't add lifecycle->ctor if other lifecycle fn ptrs are null
    registering pod components explicitly should mark lifecycle_set so that in
    translation unit B, component_info::id() doesn't override them with
    register_lifecycle_actions(true,true,true,true)
- minor: mark world::enable_tracing() as static
    it doesn't depend on world and needs to be set to get tracing info from bootstrap init

### DIFF
--- a/include/flecs/flecs.hpp
+++ b/include/flecs/flecs.hpp
@@ -715,7 +715,7 @@ public:
      *
      * @param level The tracing level.
      */
-    void enable_tracing(int level) {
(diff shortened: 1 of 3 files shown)


SanderMertens:
Thanks!

PR #253, opened 2020-09-30
Fix missing alloca() on FreeBSD

nbrk:
Unbreak build/usage on FreeBSD systems which don't have alloca.h (but have this function in stdlib.h)


SanderMertens:
Great, thanks for the PR!

PR #265, opened 2020-11-05
Changed all floats to doubles to get around floating point rounding e…

markgerl-NG:
…rrors when running multiple systems at different FPS rates.

### DIFF
--- a/src/map.c
+++ b/src/map.c
@@ -52,7 +52,7 @@ static
 int32_t get_bucket_count(
     int32_t element_count)
 {
-    return next_pow_of_2((int32_t)((double)element_count * LOAD_FACTOR));
+    return next_pow_of_2((int32_t)((float)element_count * LOAD_FACTOR));
 }
 
 static
--- a/src/map.c
+++ b/src/map.c
@@ -52,7 +52,7 @@ static
 int32_t get_bucket_count(
     int32_t element_count)
 {
-    return next_pow_of_2((int32_t)((float)element_count * LOAD_FACTOR));
+    return next_pow_of_2((int32_t)((double)element_count * LOAD_FACTOR));
 }
 
 static
--- a/src/modules/pipeline/worker.c
+++ b/src/modules/pipeline/worker.c
@@ -269,7 +269,7 @@ void ecs_set_threads(
             ecs_os_cond_free(world->sync_cond);
             ecs_os_mutex_free(world->sync_mutex);
         }
-
+        
         /* Start threads if number of threads > 1 */
         if (threads > 1) {
             world->worker_cond = ecs_os_cond_new();
--- a/src/modules/pipeline/worker.c
+++ b/src/modules/pipeline/worker.c
@@ -246,7 +246,7 @@ void ecs_workers_progress(
     }
 
     if (world->measure_frame_time) {
-        world->stats.system_time_total += (double)ecs_time_measure(&start);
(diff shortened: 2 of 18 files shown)


src/map.c
```diff
@@ -52,7 +52,7 @@ static
 int32_t get_bucket_count(
     int32_t element_count)
 {
-    return next_pow_of_2((int32_t)((float)element_count * LOAD_FACTOR));
```
SanderMertens, line 55: This is not used for FPS calculation, doesn't need to change.

src/modules/pipeline/worker.c
```diff
@@ -246,7 +246,7 @@ void ecs_workers_progress(
     }
 
     if (world->measure_frame_time) {
-        world->stats.system_time_total += (float)ecs_time_measure(&start);
+        world->stats.system_time_total += (double)ecs_time_measure(&start);
```
SanderMertens, line 249: The return type of `ecs_time_measure` is already a double, cast is not necessary.

src/world.c
```diff
@@ -661,7 +661,7 @@ void ecs_merge(
     ecs_eval_component_monitors(world);
 
     if (measure_frame_time) {
-        world->stats.merge_time_total += (float)ecs_time_measure(&t_start);
+        world->stats.merge_time_total += (double)ecs_time_measure(&t_start);
```
SanderMertens, line 664: Cast is not necessary, return type is already a double.

src/world.c
```diff
@@ -1011,13 +1011,13 @@ void stop_measure_frame(
 {
     if (world->measure_frame_time) {
         ecs_time_t t = world->frame_start_time;
-        world->stats.frame_time_total += (float)ecs_time_measure(&t);
+        world->stats.frame_time_total += (double)ecs_time_measure(&t);
```
SanderMertens, line 1014: Cast not necessary, return type is already a double.

SanderMertens (changes requested):

markgerl-NG (commented):
All seem reasonable - I've made the changes. But, I may be reading this PR wrong - you may have already made the changes. #confusing

SanderMertens:
A quick test with doubles shows that this doesn't solve the issue. This is the output for a timer that increases with intervals of 0.1, and has a timeout at 1.0:
```c
time_elapsed = 0.1000000000000000055511, timeout = 1.0000000000000000000000
time_elapsed = 0.2000000000000000111022, timeout = 1.0000000000000000000000
time_elapsed = 0.3000000000000000444089, timeout = 1.0000000000000000000000
time_elapsed = 0.4000000000000000222045, timeout = 1.0000000000000000000000
time_elapsed = 0.5000000000000000000000, timeout = 1.0000000000000000000000
time_elapsed = 0.5999999999999999777955, timeout = 1.0000000000000000000000
time_elapsed = 0.6999999999999999555911, timeout = 1.0000000000000000000000
time_elapsed = 0.7999999999999999333866, timeout = 1.0000000000000000000000
time_elapsed = 0.8999999999999999111822, timeout = 1.0000000000000000000000
time_elapsed = 0.9999999999999998889777, timeout = 1.0000000000000000000000
time_elapsed = 1.0999999999999998667732, timeout = 1.0000000000000000000000
triggered! (1.100000)
```
The system triggers at the 11th frame as the result of a floating point not having integer accuracy. This can only be addressed by using integer timeout values.

PR #275, opened 2020-11-24
Update CMakeLists.txt

Alexandre-P-J:
Solves the following CMake warning (potential issue in the future):

CMake Deprecation Warning at thirdparty/flecs/CMakeLists.txt:1 (cmake_minimum_required):
Compatibility with CMake < 2.8.12 will be removed from a future version of CMake.
Update the VERSION argument <min> value or use a ...<max> suffix to tell
CMake that the project does not need compatibility with older versions.

I'm using CMake 3.19.0


SanderMertens:
Thanks for the PR!

PR #279, opened 2020-11-28
docs: fix simple typo, entites -> entities

timgates42:
There is a small typo in examples/c/35_empty_system_signature/src/main.c, examples/cpp/35_empty_system_signature/src/main.cpp, flecs.h, include/flecs.h.

Should read `entities` rather than `entites`.


SanderMertens:
Thanks for the PR!

PR #285, opened 2020-12-17
is_number small optimization

Hexlord:
Doing 1 less isdigit call


SanderMertens:
Nice, thanks for the PR!

PR #304, opened 2021-01-22
C++ multithreading example

mcmlevi:
C++ multi threading example as the posix example didn't work on windows. This example uses the c++ standard libary for the multithreaded implmentation.


SanderMertens (approved):

SanderMertens:
Thanks a lot, looks great!

PR #309, opened 2021-02-10
optimization: remove superfluous copies of lambda functions

ikrima:
- add `noexcept` to constructors bc stl/3rd party C++ libs sometimes rely on noexcept ctor tags for optimizations (sigh, c++)
- since ctor aren't templates, && isn't treated as perfect forward reference so have to manually have two variants
- changed `call_system` function signatures to take Func&& and placed according std::move/forward
- each/action/iter_invoker calls: add explicit move and const ref ctors
- need to explicitly decay Func template parameter in std::forward invocation


SanderMertens:
Awesome, thanks for the PR!

PR #317, opened 2021-02-19
flecs.hpp: fix error #elif with no expression

SpaceIm:
This is a tiny fix in flecs.hpp, there is an #elif without expression instead of #else


SanderMertens:
Nice find, thanks!

PR #320, opened 2021-02-19
CMake: improvements (honor compiler flags, add several options, fix install and export targets)

SpaceIm:
This PR improves CMake build of flecs:

- add options to build static or shared or both (I would have loved preventing to build both static and shared -because it forces to use 2 different imported targets for shared and static- but since it's already there I can't remove this feature without eventually breaking several downstream projects)
- add an option to enable full warnings
- add an option to disable PIC on non Windows plarforms
- honor `CMAKE_*_FLAGS*` from user
- rely on `C_VISIBILITY_PRESET` property instead of -fvisibility=hidden
- remove invalid warnings options for C
- export targets in `flecs-config.cmake` to be able to consume flecs with `find_package(flecs)`, then `target_link_libraries(myapp flecs::flecs)` or `target_link_libraries(myapp flecs::flecs_static)`. These namespaced targets also work with `add_subdirectory()`.

closes https://github.com/SanderMertens/flecs/issues/319

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,6 +2,8 @@ cmake_minimum_required(VERSION 3.5)
 
 project(flecs LANGUAGES C)
 
+option(FLECS_PIC "Compile static flecs lib with position-independent-code (PIC)" ON)
+
 # include utilities for compiler options and warnings
 
 include(cmake/target_default_compile_warnings.cmake)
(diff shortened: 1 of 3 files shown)


rawbby (approved):
@SpaceIm That is my fault :/
Copied some of the CMake code from a project of mine where I wanted specific compiler flags to be set. Some CMakes predefined flags where incompatible with mine so I decided to explicitly (un)set them.
Now I see why this is not a good idea.

rawbby (approved):

SanderMertens:
Oops, just realized that you pushed fixes after I merged master into the branch @SpaceIm. Feel free to reopen the issue / create a new PR if there were more things you'd like to add.

SpaceIm:
@SanderMertens LGTM, the last commit since to be in master ;)

SanderMertens:
Great, thanks for the PR! @rawbby thanks for reviewing :)

PR #329, opened 2021-03-04
Improving bazel build.

mason-bially:
Using `includes` over `copts` for public includes. This fixes the main problem of being able to be correctly depended on by other projects. I also am correcting the build to include private headers, and to be lower maintenance in general.

I also took the liberty of adding the test projects to bazel. One thing we could do is shard the `:api` test group. To do this I had to add the bake test framework via a `new` tool. It might be worth moving the inline build file for `bake` to the appropriate repository. On the other hand all I am pulling is just some pieces of it.

I also improved the example by making it a separate workspace with a local reference. One improvement here is to include an actual proper git reference, but a good version of the bazel build would have to be committed *first* before that bazel reference could be added.


mason-bially:
Responding to feedback from discord.

Also moved build files around to clean up the core files.

mason-bially:
- `bazel test ...` should run all tests.
- `bazel build flecs` should build properly.
- And from `/examples/build/bazel` `bazel run example` should work properly.

SanderMertens:
Nice! Thanks for also fixing the addon macro's :)

PR #358, opened 2021-03-24
New relationship API (deprecates traits, and a lot more)

### LINKED ISSUE #322: Remove overhead of STL
**Describe the problem you are trying to solve.**
Currently the C++ API includes a number of headers that can significantly increase compile time (string, sstream) without providing much benefit.

This image shows the (compile time) overhead (provided by @ikrima):
header_overhead

**Describe the solution you'd like**
Remove unnecessary STL dependencies.

### LINKED ISSUE #345: Update API & query parser to new entity-relation terminology/notation
**Describe the problem you are trying to solve.**
With the upcoming new query implementation (#295) the primary use case for traits will be as a way to store relationships between entities. The terminology used by the current API does not make this obvious or intuitive.

Additionally, there are a few inconsistencies between the C and C++ API that need to be addressed (the C API uses the reverse order of specifying traits as the C++ API).

**Describe the solution you'd like**
Modify the API & query parser to use terminology/notation as described here:
https://github.com/SanderMertens/flecs/discussions/344

SanderMertens:
This PR introduces the new relationship API as described here: https://github.com/SanderMertens/flecs/discussions/344. That is not the only thing that changed though. The API has been reduced in size, with clearer naming and less verbose functions. The existing APIs still work, but are deprecated.

After upgrading to the new API existing applications will still work as expected. I have introduced two new macro's however that control how calling deprecated functions is treated:

`FLECS_NO_DEPRECATED_WARNINGS` (defined by default): This suppresses warnings about usage of deprecated functions. Undefining this macro will likely throw a lot of warnings for an unmodified application, but the application will still work as usual.

`FLECS_DEPRECATED` (defined by default): This controls the "deprecated" addon. All deprecated functions are moved to an addon that can be excluded from a build. Undefining this macro will cause unmodified code to break, as deprecated functions will no longer be available.

The `FLECS_NO_DEPRECATED_WARNINGS` macro will become undefined by default in the 2.5 release (the next release is 2.4). After that unmodified code that relies on deprecated functions will start throwing warnings. Applications will be able to redefine the macro if they are not yet ready to upgrade.

The deprecated functions will be definitively removed from the source code with the next major release (v3).

In addition to the relationship API changes, the following things have been changed as well:

- A new core type `ecs_id_t` has been introduced. This type is used in function signatures whenever a function expects an identifier that can be either an entity, a pair (relationship) or a role annotated identifier. This makes it clearer which arguments are expected to be valid entities, and which arguments can be either entity (component) ids or pair ids.

- Functions with postfix `_entity` or `_w_entity` have been renamed to `_id` or `_w_id`.

- deprecation of `_type` functions. these will be partially replaced with more powerful filter-based alternatives

- deprecation of ECS_ENTITY, ECS_TYPE,  ECS_PREFAB, these will be partially replaced with more powerful alternatives that are not macro-based

- The C++ API no longer has a dependency on STL (save for type_traits)

- The code now compiles warning free with stricter warning settings

- improved lifecycle handling of pair ids and API primitives to easier synchronize ids between data sources

- a fix for a nasty memory corruption bug

closes #345
closes #322


CMakeLists.txt
```diff
@@ -20,77 +20,9 @@ include(cmake/target_default_compile_options.cmake)
 # collect all include and source files
 # (use the manual list to improve ide support)
```
Hexlord, line 21: No longer true :)

examples/c/38_delta_system_time/src/main.c
```diff
@@ -1,15 +1,15 @@
 #include <delta_system_time.h>
 
 void Period05(ecs_iter_t *it) {
-    printf("t = 0.5, elapsed time = %f\n", it->delta_system_time);
+    printf("t = 0.5, elapsed time = %f\n", (double)it->delta_system_time);
```
Hexlord, line 4: What is the point of this cast?

examples/cpp/10_inheritance/src/main.cpp
```diff
@@ -71,17 +69,17 @@ int main(int argc, char *argv[]) {
 
     /* Create entities which share the Mass component from a base */
     ecs.entity("MyInstance1")
-        .add_instanceof(LightEntity)
+        .add(flecs::IsA, LightEntity)
```
Hexlord, line 72: Would love for some better naming, like a method .instantiate(LightEntity), love the removal of add_childof though!

include/flecs/cpp/system.hpp
```diff
@@ -297,10 +302,10 @@ class system final : public entity {
             is_each = false;
         }
 
-        std::string signature = build_signature(is_each);
+        flecs::string signature = build_signature(is_each);
 
         if (!signature.length()) {
-            signature = "0";
+            signature = flecs::string_view("0");
```
Hexlord, line 308: flecs::string assignment operator for (const char*) might be handy not to do such boilerplate

Hexlord (approved):
Blasted through it, looks fine (5% of it that I've actually looked at)

examples/c/38_delta_system_time/src/main.c
SanderMertens: I added a double-promotion warning to the build which ensures that you can't implicitly upcast from a float to a double. I use the examples to verify that you can use the API without warnings, but that means that they also have to adhere to the strictest warning settings.

It's a bit of an annoying warning, because it will trigger each time you try to pass a float to printf, but still that's better than having flecs compile with warnings.

examples/cpp/10_inheritance/src/main.cpp
SanderMertens: I want to reduce the API footprint, so I think I'll stick to this :) Might take a bit to get used to, but in the context of the new relationship/query framework this makes it clearer how instancing is implemented

include/flecs/cpp/system.hpp
SanderMertens: I actually explicitly did not want this. What happens here is I'm assigning a `string_view` to a `string` instance, which causes the `string` to hold a non-owning pointer. No copy is created.

The string classes I've built are for internal usage only, and I don't want the copy-by-default behavior of std::string (it's wasteful for most functions). That means the resulting code will be a bit more verbose, but it is unambiguous what is happening from looking at the code.

CMakeLists.txt
SanderMertens: That's true!

Hexlord: Oh, ok, sorry, I was thinking you meant You are using the manual list, but you mean it for the user to replace the list with the manual one, yes it is true

include/flecs/cpp/system.hpp
Hexlord: Oh wow, that is smart! Nice

PR #371, opened 2021-04-07
Handle a null world in is_valid and is_alive

shanehyde:
If `m_world` is null on a `flecs::entity` then `is_valid()` asserts, which doesn't make any sense on a validity check.  This small PR just returns false for `is_valid()` and `is_alive()` if the `m_world` value is null


SanderMertens:
Good point. It is a bit inconsistent with the C functions, but it makes sense from the API perspective. Thanks for the PR!

PR #372, opened 2021-04-12
Add build commands to examples for reference

aganm:
I added build commands to the examples for my own convenience 😄 

Preview the changes here [https://github.com/aganm/flecs/tree/buildexample#example](https://github.com/aganm/flecs/tree/buildexample#example)


SanderMertens:
Good idea to add these, but I think it's a bit cleaner to keep the build examples under the build section :) I'll make sure to add them!

PR #431, opened 2021-06-09
Fix duplicate cond

aganm:
Little fix for this warning
```
flecs.c: In function ‘ecs_filter_str’:
flecs.c:17107:36: warning: duplicated ‘if’ condition [-Wduplicated-cond]
17107 |             } else if (term->inout == EcsInOut) {            
      |                        ~~~~~~~~~~~~^~~~~~~~~~~
/home/aganm/Projects/gow/lib/flecs/flecs.c:17105:36: note: previously used here
17105 |             } else if (term->inout == EcsInOut) {

```


SanderMertens:
Nice catch!

PR #432, opened 2021-06-13
test/api/SystemMisc: add assert

randy408:


SanderMertens:
@randy408 could you provide the reason for the change? Given this code:

```c
    ECS_COMPONENT(world, Position);
    ECS_COMPONENT(world, Velocity);
    ECS_ENTITY(world, f, Velocity);

    ecs_query_t *q = ecs_query_new(world, "Position, f:Velocity");

    ecs_entity_t e = ecs_new(world, Position);
    ecs_add(world, e, Velocity);
```

only `e` matches the query, as it is the only entity that has `Position`. The query only requests `Velocity` explicitly from `f`.

randy408:
> only `e` matches the query

Right, there doesn't seem to be a positive test for the `entity:Component` syntax, should I extend the test case with with a second query?

randy408:
It didn't occur to me that `entity:Component` will add a shared column and an iterator with only shared columns will have an `it.count` of 0.

When `entity:Component` is not found it asserts, closing.

Issue #389, opened 2021-05-02
System declaration order not respected after deferred creation with multiple threads


Hexlord:
**Describe the bug**
After the deferred spawn of an entity with a component, systems involving the component are reordered, but only when using multiple threads AND spawning _FIRST match of the query_ in deferred mode. 
Adding `ECS.entity().set<CColoredStaticShape>({}).destruct();` (so creating the component and immediately destructing it - I guess thus forcing some query/pipeline rebuilds) after the last system declaration fixes the order (for the whole run of the program), it becomes Tick -> **S1 -> S1 -> S2** (so this is the workaround). 
Thus my guess is that deferred spawn component, query and pipeline interaction might be buggy.

**To Reproduce**
```c
stdcpp_set_os_api();
flecs::world World;
World.set_threads(8);

    struct CColoredStaticShape {
        int Data;
    };
    struct CColoredStaticShapeChange {
        int Data;
    };
    World.use<CColoredStaticShape>();
    World.use<CColoredStaticShapeChange>();
    World.system(nullptr, "").iter([](iter&Iter){
      static int Delay = 10;
      if(--Delay == 0) {
          printf("Spawning...");
          Iter.world().entity().set<CColoredStaticShape>({});
      }
    });
    World.system(nullptr, "[in] CColoredStaticShape, [out]:CColoredStaticShapeChange").iter([](iter& Iter){
      printf("S1");
      for (auto Index : Iter) {
          if (rand() % 2 == 0) {
              Iter.entity(Index).set<CColoredStaticShapeChange>({});
          }
      }
    });

    World.system(nullptr, "[inout] :CColoredStaticShapeChange").iter([](iter&){
      printf("S2");
    });
    // Uncomment below to activate the magical Workaround.
    // World.entity().set<CColoredStaticShape>({}).destruct();
    while(true) {
        static int i = 0;
        printf("Tick %i", ++i);
        World.progress(0.1f);
        std::this_thread::sleep_for(std::chrono::milliseconds(100));
    }

```

**Expected behavior**
After the spawn, debug output should be like: Tick 11 -> **S1 -> S2**, so that S1 happens before S2 within the same frame.

**Actual behavior**
After the spawn, debug output is like: Tick 11 -> **S2 -> S1**. S2 happens before S1 within the same frame.

PR #442, opened 2021-07-10
#389 Delaying system activation/deactivation

Hexlord:
In multi-threaded environment system order could be messed up due to mid-frame system activation (result of table activation). This tries to fix this by delaying the activation through additional tags EcsDelayedActivation and EcsDelayedDeactivation.
See #389 for more details.

Code is surely terrible, it is just to illustrate the idea, hope @SanderMertens fixes it up to match the `flecs` ideology.

It works for code in the issue - the order is now respected.


SanderMertens:
Closing this PR, as discussed on Discord the solution should take a somewhat different approach. Codebase has also diverged quite a bit with thew new v3 work.

Issue #461, opened 2021-08-03
entity and entity_view should be implicitly default constructible


Hexlord:
**Describe the problem you are trying to solve.**
Given 
```c
struct Comp {
    flecs::entity_view Entity;
}
```
We are able to write
```c
world.entity().add<Comp>();
```
but not
```c
world.entity().set<Comp>({});
```
because `Comp` construction from initializer list `{}` is considered implicit, and the default constructors of `entity` and `entity_view` are currently marked explicit.

**Describe the solution you'd like**
Make both default constructors implicit.

SanderMertens:
Merged it!

PR #462, opened 2021-08-03
#461 make entity and entity_view implicitly default constructible

Hexlord:
Made `entity` and `entity_view` default constructors implicit.

See #461 for more details.


SanderMertens:
Thx for the PR!

PR #479, opened 2021-09-03
Fixed issue with on set systems for pair types

BHolman-LBI:
Addressed issue with on set systems with pairs when matching multiple columns on a single table only triggering for the last column.


BHolman-LBI:
I think I need an LTS branch to open this PR

PR #484, opened 2021-09-13
Update hash.c to allow valgrind to always be defined, and support address sanitizer

logankaser:
Fixes building when user has defined `VALGRIND` and tries to build a debug build, for
compilers that do not allow `VALGRIND` to be defined twice. Also adds support for address sanitizer's
most common warding define.


SanderMertens:

PR #513, opened 2021-10-12
Correct documentation for ecs_set_symbol

jemc:
The prior doc comment says that it sets the given string as the `EcsName` component, but in reality, it uses the `EcsSymbol` component.

This PR fixes the doc comment to match that reality.


SanderMertens:
@jemc @randy408 is right, the name/symbol values are stored in pairs with `EcsIdentifier`. I updated the out of date comments/docs, thanks for reporting!

Issue #124, opened 2020-01-06
Internal improvements


SanderMertens:
Catch-all issue for internal improvements not directly related to a feature.

PR #535, opened 2021-11-08
#124 change query case terms to a (switch, case) pair

SanderMertens:
Queries for a case now need to provide both the switch and case. See this discussion for more details: https://github.com/SanderMertens/flecs/discussions/466#discussioncomment-1605624


BHolman-LBI (approved):
Looks good -- I don't see anything obviously incorrect.

PR #539, opened 2021-11-11
[doc] Add missing for loop in System C++ example in Quickstart

ZeroErrors:
Was reading through the Quickstart and discovered a minor bug in the System C++ example.


SanderMertens:
Nice catch, thanks for the fix!

PR #556, opened 2021-11-22
Fix typo in "relations" docs

StickyFingies:
A lambda callback was missing its open parentheses.


SanderMertens:
Nice catch, thanks for the PR!

Issue #558, opened 2021-11-27
Retrieve InOut-ness from the const-ness of term()'s T argument


Hexlord:
**Describe the problem you are trying to solve.**
Writing less code.

**Describe the solution you'd like**
If `.term()` is used like `.term<const T>()` we would like `.inout(flecs::In)` to be automatically implied for said term.

SanderMertens:
It already works like this!

PR #559, opened 2021-11-27
#558 Retriving InOut type from term argument type

Hexlord:
Added possible implementation of implicit InOut derived from term's T type like in Observer, lets see if it passes tests.
implements #558


SanderMertens:
Thanks! 🙏

PR #566, opened 2021-11-30
Cleaning up the BAZEL build to work with the new code, and on Windows.

mason-bially:
Updating the bazel build files to support version 3. I can fix up the example once it's officially released.

So some issues with this build.

I can't put the BUILD file in just the root. This breaks Windows/MSVC due to how it likes to resolve `#include` dependencies (because there is a `flecs.h` in the root folder). I also can't put an alias there to enable `@flecs` resolving correctly because for whatever reason building `//src:flecs` from the root folder also breaks it. Something is wrong in bazel's sandboxing in windows for this specific situation (code at the root).

Some tests are not passing on windows. Notably these won't even build because MSVC does not support C99.

```
test/api/src/Filter.c(789): error C2059: syntax error: '}'
test/api/src/Filter.c(822): error C2059: syntax error: '}'
test/api/src/Filter.c(859): error C2059: syntax error: '}'
test/api/src/Filter.c(984): error C2059: syntax error: '}'
test/api/src/Filter.c(1044): error C2059: syntax error: '}'

test/api/src/OnDelete.c(717): error C2057: expected constant expression
test/api/src/OnDelete.c(717): error C2466: cannot allocate an array of constant size 0
test/api/src/OnDelete.c(717): error C2133: 'r': unknown size
test/api/src/OnDelete.c(748): error C2057: expected constant expression
test/api/src/OnDelete.c(748): error C2466: cannot allocate an array of constant size 0
test/api/src/OnDelete.c(748): error C2133: 'r': unknown size
test/api/src/OnDelete.c(1025): error C2057: expected constant expression
test/api/src/OnDelete.c(1025): error C2466: cannot allocate an array of constant size 0
test/api/src/OnDelete.c(1025): error C2133: 'objects': unknown size

test/api/src/Rules.c(32): error C2059: syntax error: '}'
test/api/src/Rules.c(4227): error C2059: syntax error: '}'

test/api/src/Entity.c(718): error C2059: syntax error: '}'
test/api/src/Entity.c(757): error C2059: syntax error: '}'
```


mason-bially:
@SanderMertens I'm not sure why the builds are failing, appears to be a bake issue? I don't think I'm effecting it.

This should conform closer to what you want the bazel build to look like.

logankaser:
I've been using something a lot like this:
```python
config_setting(
    name = "debug",
    values = {"compilation_mode": "dbg"},
)

cc_library(
    name = "flecs",
    srcs = glob([
        "src/**/*.c",
        "src/**/*.h",
    ]),
    hdrs = glob([
        "include/flecs/private/*.h",
        "include/flecs/addons/**/*.h",
        "include/flecs/addons/**/*.hpp",
        "include/*.h",
        "include/flecs/*.h",
    ]) + [
        "flecs.h",
    ],
    defines = select({
        ":debug": [
            "FLECS_DEBUG",
            "FLECS_LOG_2",
        ],
        "//conditions:default": [
            "FLECS_NDEBUG",
            "FLECS_LOG_1",
        ],
    }),
    includes = ["include"],
    linkopts = ["-lpthread"],
    textual_hdrs = glob([
        "src/**/*.inl",
        "include/**/*.inl",
    ]),
)
```
You are welcome to take inspiration from it / copy it.
It has some advantages in terms of keeping non-public parts of the API non-public.
The debug support is also nice.

mason-bially:
I don't think that build is going to work on windows against MSVC. Which was one of the goals here.

PR #605, opened 2022-01-11
Fixing C4706 MSVC warning: assignment within conditional expression

jtferson:
When compiling with MSVC in Visual Studio or Rider for Unreal it complains about 'assignment within conditional expression' (C4706 warning):
`while ((ptr = strstr(ptr + 1, "struct "))) { }`

To fix this issue we need to explicitly test the result of the assignment:
`while ((ptr = strstr(ptr + 1, "struct ")) != 0) { }`


SanderMertens:
Thanks for the PR! @Lexxicon also ran into this issue.

PR #631, opened 2022-01-29
Verify omitted members break reflection

Lexxicon:
Added test checking if omitting a member from a registered component breaks things


SanderMertens:
Fixed!

PR #649, opened 2022-02-15
Install *.inl files with headers.

photex:
Howdy!

I have been working with flecs via conan and so I've started using CMake again. After setting up a recipe for flecs v3 I noticed that *.inl files don't get installed.


SanderMertens:
Thanks for the PR!

PR #654, opened 2022-02-22
Updated Pipeline tags in Quickstart.md

MewSoul:
## Change Log
- Renamed `*PostStore` to `*OnStore` for both C/C++ doc
- Removed `flecs::PostValidate` from C++ doc
- Moved `flecs::PostUpdate` after `flecs::OnValidate` in C++ doc

### Note

@SanderMertens Is it intended that the phases `PreFrame` and `PostFrame` are not mentioned? Or should they be added to the documentation?

### DIFF
--- a/docs/Quickstart.md
+++ b/docs/Quickstart.md
@@ -673,18 +673,17 @@ EcsOnUpdate
 EcsOnValidate
 EcsPostUpdate
 EcsPreStore
-EcsPostStore
+EcsOnStore
 ```
 ```cpp
 flecs::OnLoad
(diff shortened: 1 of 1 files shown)


SanderMertens:
> Is it intended that the phases PreFrame and PostFrame are not mentioned? Or should they be added to the documentation?

That's right! The `PreFrame` and `PostFrame` are builtin tags that are used to run internal systems, like the ones that progress timers. Systems that run before `PreFrame` or after `PostFrame` may not work as expected.

PR #674, opened 2022-03-15
#673 Prevent rare deadlock on the last worker thread

### LINKED ISSUE #673: Rare deadlock on the last worker thread
**Describe the bug**
Sometimes during world destruction one of the worker threads never joins - it just hangs on worker_cond.

**To Reproduce**
```c
for(uint32 Index = 0; Index < 10000; ++Index) {
    flecs::world ECS;
    ECS.set_threads(8);
    ECS.system("No Staging System").no_staging().iter([](iter &Iter) {
        auto ECS = Iter.world();
        ECS.defer_end();
        assert(!ECS.is_deferred());
        ECS.defer_begin();
    });
    ECS.progress();
    ECS.progress();
    printf("%u\n", Index);
}
```
This never reaches 10000, it deadlocks beforehand for me.

I am on Linux Mint 20.2 Cinnamon Intel© Xeon© CPU E3-1231 v3 @ 3.40GHz × 4

Hexlord:
Sometimes last worker just deadlocks on the worker_cond, it seems that this change fixes #673 .

I am on Linux Mint 20.2 Cinnamon Intel© Xeon© CPU E3-1231 v3 @ 3.40GHz × 4

### DIFF
--- a/src/addons/pipeline/worker.c
+++ b/src/addons/pipeline/worker.c
@@ -91,9 +91,12 @@ void sync_worker(
         /* Only signal main thread when all threads are waiting */
         ecs_os_cond_signal(world->sync_cond);
     }
+    
+    if (!world->quit_workers) {
+        /* Wait until main thread signals that thread can continue */
+        ecs_os_cond_wait(world->worker_cond, world->sync_mutex);
+    }
 
-    /* Wait until main thread signals that thread can continue */
-    ecs_os_cond_wait(world->worker_cond, world->sync_mutex);
     ecs_os_mutex_unlock(world->sync_mutex);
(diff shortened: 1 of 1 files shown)


SanderMertens:
Hm, can you describe the scenario that this PR addresses? The `world->quit_workers` variable should never be true while `sync_worker`, as this function is only called while the pipeline is running. If this solves your deadlock it suggests that you call `ecs_stop_threads` while the pipeline is running, which is not allowed (doesn't look like it would assert though- that's something I should add).

Hexlord:
This does not fix it. I'll add repro to the issue

PR #708, opened 2022-04-23
res_index modulation

Hexlord:
Better thread utilization for tables with entity count lower than worker count through resource index modulation


SanderMertens:
I'll try to take a look at the failing tests soon. If it's just ordering issues this PR should be good to go

PR #713, opened 2022-05-03
Issue 677 - Use Windows os_api_impl on non-MSVC Windows Targets

Jared-Miller:
I manually modified `flecs.c` since bake-amalgamate doesn't currently work on Windows.


SanderMertens:
Thanks for the PR!

PR #717, opened 2022-05-15
Revert "Expose m_world in flecs::query"

Hexlord:
This did not help me in my scenario - comparing static storage query's world to argument world was succeeding, but in fact it was just world pointers being the same, while in fact the world has changed, so I ended up storing an unique ID of the world inside a singleton and comparing it instead, feel free to revert this extraneous interface


SanderMertens:
I think being able to get the world() from a query is fine, so I'll just leave it in.

PR #720, opened 2022-05-25
Map linked list block allocator

Hexlord:
Switch map from dynamic arrays to linked list with block allocator.


SanderMertens:
Nice performance boost, thanks!

Issue #731, opened 2022-06-06
Faster table sorting for queries


bazhenovc:
# The problem
Following our Discord discussion:

Currently `ecs_order_by_action_t` is defined as a function pointer that operates on two individual elements, then there is a `sort_tables()` function that sorts the data and calls this function by pointer.

In this case the compiler cannot inline the comparator function and will always be calling it by pointer, so it's basically an equivalent of a C++ virtual function call for each element - it is very slow.

# Proposed solution
Define `ecs_order_by_action_t` as a function pointer that sorts the entire table, then define `FLECS_DECLARE_ORDER_BY_ACTION` helper macro that produces this sorting function. The macro should be similar to how `datastructures/qsort` works.

For example:

```c
FLECS_DECLARE_ORDER_BY_ACTION(MyCustomOrderingFunction, CompareTwoElements);

ecs_query_desc_t desc = {};
desc.order_by = MyCustomOrderingFunction;
```

## Helper macro
Macro `FLECS_DECLARE_ORDER_BY_ACTION` will define the entire sorting function named `MyCustomOrderingFunction` that will call `CompareTwoElements` directly, so that the compiler will see all the definitions and properly optimize and inline everything.

## Sort function signature

There currently is a minor issue with `sort_tables()` - it accesses the members of `ecs_table_t`, which is an internal opaque type. Proposed solution: define `ecs_order_by_action_t` like this and expose `ecs_column_t` to the user:

```c
void (*ecs_order_by_action_t)(
  int32_t count,                // number of elements to sort
  ecs_entity_t *entities,       // entity ids
  ecs_column_t *columns,        // component arrays
  ecs_type_info_t **type_info,  // array with type info structs
  int32_t column_count,         // total number of columns
  int32_t column_to_sort)       // column with component to sort
```

The signature should be completely irrelevant for the user because the helper macro will take care of it. However it will have binary compatibility implications.

## C++ bindings

The C++ bindings should use a templated version of `ecs_order_by_action_t`. Comparing two elements can be implemented either using `std::less` that the user will provide specializations for the component type or by passing a sorting closure by value.

With C++ bindings it is currently unclear how to handle swapping and iteration of the arrays, the C interface needs to be finalized before implementing it. I don't see any fundamental issues there, will figure it out once I get there.

SanderMertens:
Looks good! Few thoughts:

[...]

PR #734, opened 2022-06-07
Faster table sorting for queries.

bazhenovc:
This PR implements #731 

Heavily work in progress, the implementation is going to take some time so this PR is for additional discussions and tracking progress.

I will squash commits later if needed.


bazhenovc:
@SanderMertens I had a few questions along the way 🙂 

First question - it turns out that the table sorting code is calling `flecs_table_swap()` internally, which is private. I made a temporary workaround for the time being. What would be the best way to tackle that?

Second question/suggestion - maybe it would be better to define something like `ecs_order_by_action_t` like this?

```c
typedef struct ecs_order_by_action_t {
ecs_compare_component_action_t compare;
ecs_sort_table_action_t sort_table;
};
```

And have the query to hold this struct instead of 2 pointers? The reason is that even if we have a `sort_table` implemented, there are still calls to `compare()` on individual elements in `build_sorted_table_range()` so it looks like both implementations will be needed. This shouldn't impact the C++ bindings.

SanderMertens:
> First question - it turns out that the table sorting code is calling flecs_table_swap() internally, which is private. I made a temporary workaround for the time being. What would be the best way to tackle that?

Right, the challenge here is that the swap implementation is not trivial as it needs to correctly invoke component (move) hooks if they're registered, and allocate temporary storage for the swap. I think for this to work correctly the entire implementation of the swap would have to be done by the macro.

This does significantly complicate the interface, as not everyone might want to use the macro (most obvious example would be language bindings, runtime components). It could be worth exploring a design (maybe as a stepping stone) where instead of calling `flecs_table_swap` many times, a new `flecs_table_sort` function is introduced that accepts a comparator function, this would be easier to do and should also show improvements in performance.

> Second question/suggestion - maybe it would be better to define something like ecs_order_by_action_t like this?
For the first iteration I'd keep this private, as it's tightly coupled with internals. I'd like to keep the option open to change the way table slices are sorted internally without breaking the API.

bazhenovc:
`flecs_table_swap()` does not depend on the element type and doesn't need a comparator, I don't think I fully understand why it needs to be implemented by the macro?

SanderMertens:
Strictly speaking it's not necessary, but it would be beneficial for performance if the order_by function could order multiple arrays inline. Consider all the code `flecs_table_swap` is executing before the _actual_ swap happens: https://github.com/SanderMertens/flecs/blob/master/src/table.c#L1957

Most of this is pure overhead that would only need to happen once if the sort itself operates directly on the table data.

bazhenovc:
Sounds good, I'll start looking into `flecs_table_sort()` implementation then.

bazhenovc:
I had a deeper look at `flecs_table_swap()` and I think that the majority of work that it is doing is actually per-element and cannot happen per-table once.

Here's a breakdown:

```c
void flecs_table_swap(
    ecs_world_t *world,
    ecs_table_t *table,
    ecs_data_t *data,
    int32_t row_1,
    int32_t row_2)
{    
    (void)world;

    ecs_assert(!table->lock, ECS_LOCKED_STORAGE, NULL);
    ecs_assert(data != NULL, ECS_INTERNAL_ERROR, NULL);
    ecs_assert(row_1 >= 0, ECS_INTERNAL_ERROR, NULL);
    ecs_assert(row_2 >= 0, ECS_INTERNAL_ERROR, NULL);
```

These 4 asserts are fine I think, maybe checking `!table->lock` could be done once per table, but I don't see any harm here. Checking `data`, `row_1` and `row_2` are specific to the two elements being swapped, so it cannot be done once per table.

```c
    check_table_sanity(table);
    
    if (row_1 == row_2) {
        return;
    }

    /* If the table is monitored indicate that there has been a change */
    mark_table_dirty(world, table, 0);
```

`check_table_sanity()` is only enabled when `FLECS_SANITIZE` is defined, again it could probably be done once per table, but I don't see any harm from doing it here - it might in fact be beneficial if the data is corrupted after swapping? Otherwise it could just be moved higher on the call stack very easily, since it doesn't depend on any of the inputs.

`mark_table_dirty()` I believe is necessary here - it will increment an internal version number when elements are swapped. Is not using any inputs, but it still should mark the table dirty when `row_1 != row_2` because the table will be dirty.

```c
    ecs_entity_t *entities = data->entities.array;
    ecs_entity_t e1 = entities[row_1];
    ecs_entity_t e2 = entities[row_2];

    ecs_record_t **records = data->records.array;
    ecs_record_t *record_ptr_1 = records[row_1];
    ecs_record_t *record_ptr_2 = records[row_2];

    ecs_assert(record_ptr_1 != NULL, ECS_INTERNAL_ERROR, NULL);
    ecs_assert(record_ptr_2 != NULL, ECS_INTERNAL_ERROR, NULL);

    /* Keep track of whether entity is watched */
    uint32_t flags_1 = ECS_RECORD_TO_ROW_FLAGS(record_ptr_1->row);
    uint32_t flags_2 = ECS_RECORD_TO_ROW_FLAGS(record_ptr_2->row);
```

All this stuff is specific to the two elements being swapped - `e1`, `e2`, `record_ptr_1`, `record_ptr_2` are fetched using `row_1` and `row_2`, this cannot be done once per table. Same deal with `flags_1` and `flags_2`.

```c
    /* Swap entities & records */
    entities[row_1] = e2;
    entities[row_2] = e1;
    record_ptr_1->row = ECS_ROW_TO_RECORD(row_2, flags_1);
    record_ptr_2->row = ECS_ROW_TO_RECORD(row_1, flags_2);
    records[row_1] = record_ptr_2;
    records[row_2] = record_ptr_1;

    swap_switch_columns(table, data, row_1, row_2);
    swap_bitset_columns(table, data, row_1, row_2);  

    ecs_column_t *columns = data->columns;
    if (!columns) {
        check_table_sanity(table);
        return;
    }

    ecs_type_info_t **type_info = table->type_info;

    /* Swap columns */
    int32_t i, column_count = table->storage_count;
    for (i = 0; i < column_count; i ++) {
        ecs_type_info_t *ti = type_info[i];
        int32_t size = ti->size;
        ecs_assert(size != 0, ECS_INTERNAL_ERROR, NULL);

        void *ptr = columns[i].array;
        void *tmp = ecs_os_alloca(size);

        void *el_1 = ECS_ELEM(ptr, size, row_1);
        void *el_2 = ECS_ELEM(ptr, size, row_2);

        ecs_os_memcpy(tmp, el_1, size);
        ecs_os_memcpy(el_1, el_2, size);
        ecs_os_memcpy(el_2, tmp, size);
    }

    check_table_sanity(table);
```

This is the rest of the function which is the actual swapping, again it doesn't seem like anything from this code can be done per-table because it depends on the `row_1` and `row_2` inputs?

The final `check_table_sanity()` - same idea, I believe that it makes sense to do it here the second time.

So in the end I think that it would be better to keep this function as-is, because there's nothing here that can be ran once per table. And since it does not use any function pointers it should be reasonably fast, when LTO is enabled the compiler will be able to inline everything.

SanderMertens:
Agree with most things, a few notes:
- `mark_table_dirty(world, table, 0)` only needs to be called once per sort, not once per swap
- the temporary swap storage could be done once (the alloca calls)
- each `->` on the table is a potential load which has overhead

You're probably right though, LTO should be able to optimize the latter two points.

> So in the end I think that it would be better to keep this function as-is

I'm OK with this. The proposed interface doesn't expose internals beyond `ecs_column_t`, which keeps the option open to optimize this further down the line if we wanted to.

bazhenovc:
> mark_table_dirty(world, table, 0) only needs to be called once per sort, not once per swap
I'll do that, no problem.

> the temporary swap storage could be done once (the alloca calls)
yep good point, also doing it in a loop is a potential stack overflow. I'll fix that too.

bazhenovc:
Actually that alloca cannot be moved outside of the loop 😄 It needs the size from column type info.

SanderMertens:
> Actually that alloca cannot be moved outside of the loop

You could (in theory- not saying it needs to be done now) do a single loop over the columns before the sort to find out the max size of all the columns, then do a single alloca of which you could reuse the result for all swaps.

bazhenovc:
@SanderMertens I think I'm happy with how it looks. Do you have any additional suggestions/requirements?

bazhenovc:
Ah snap, some tests didn't compile. I'll fix that tomorrow evening, sorry about that 🙂

SanderMertens:
No worries! I need to figure out why Github keeps asking me to give permission to run CI 🤔 

Few remarks:
- Let's for now keep the `ecs_order_by_action_t` name, to minimize breaking changes (I do agree with the intent)
- More a maintainability issue than anything else: it would be nice if the existing sort (the one that accepts a callback) could be implemented using the new table sort. That way the codebase doesn't contain the same sort implementation twice.

I'll do a more in depth review tomorrow. Looking good so far, appreciate the effort!

bazhenovc:
> Let's for now keep the ecs_order_by_action_t name, to minimize breaking changes (I do agree with the intent)

Sounds good, I'll rename it back.

> More a maintainability issue than anything else: it would be nice if the existing sort (the one that accepts a callback) could be implemented using the new table sort. That way the codebase doesn't contain the same sort implementation twice.

Good point, I'll look into that.

bazhenovc:
@SanderMertens all done 🙂

include/flecs.h
```diff
@@ -203,6 +203,9 @@ typedef struct ecs_ref_t ecs_ref_t;
 /** Type information */
 typedef struct ecs_type_info_t ecs_type_info_t;
 
+/** Table data */
+typedef struct ecs_data_t ecs_data_t;
```
SanderMertens, line 207: I think this type can remain private (see below). It is likely to change in future releases as there are a couple of improvements on the roadmap that may change how it relates to a table.

include/flecs/addons/flecs_c.h
```diff
@@ -102,6 +102,117 @@
...
+                el = ECS_ELEM(ptr, elem_size, i); \
+            } while ( compare_fn(entities[i], el, pivot_e, pivot) < 0); \
+            do { \
+                j --; \
+                el = ECS_ELEM(ptr, elem_size, j); \
+            } while ( compare_fn(entities[j], el, pivot_e, pivot) > 0); \
+            if (i >= j) { \
+                return j; \
+            } \
+            ecs_table_swap_rows(world, table, data, i, j); \
```
SanderMertens, line 162: I don't think there are any places in the code where `ecs_table_swap_rows` is used with a `data` argument that's different from that of the table. I'd prefer removing this argument from the function, so that it can also be removed from the sort signature.

src/query.c
```diff
@@ -882,84 +882,15 @@ bool match_table(
...
-    }
-
-    int32_t p = qsort_partition(
-        world, table, data, entities, ptr, size, lo, hi, compare);
-
-    qsort_array(world, table, data, entities, ptr, size, lo, p, compare);
-
-    qsort_array(world, table, data, entities, ptr, size, p + 1, hi, compare); 
-}
+ECS_SORT_TABLE_WITH_COMPARE(_, sort_table_generic, order_by, static)
```
SanderMertens, line 885: Nice!

src/table.c
```diff
@@ -2002,15 +2002,23 @@ void flecs_table_swap(
 
     ecs_type_info_t **type_info = table->type_info;
 
+    /* Find the maximum size of column elements
+     * and allocate a temporary buffer for swapping */
+    int32_t i, temp_buffer_size = sizeof(uint64_t), column_count = table->storage_count;
```
SanderMertens, line 2007: Can you replace `sizeof` with `ECS_SIZEOF`? This should've thrown a warning, as it assigns an unsigned integer to a signed integer.
```c
temp_buffer_size = sizeof(uint64_t)
```

SanderMertens (changes requested):

SanderMertens:
Just finished the review, had a few minor changes, otherwise looks good for merge :)

bazhenovc:
All done, hopefully there are no more CI build issues 😄

src/query.c
bazhenovc: Thanks!

Counts as "changes requested" for some reason, but not a big deal I guess 😄

SanderMertens (approved):

SanderMertens:
Merged it! Thanks for contributing 🙏

PR #746, opened 2022-06-17
Fix win clang linker errors (link winsock explicitly)

BlackxSnow:
Mingw clang throws linker errors from `addons/http.c` due to not implicitly linking the winsock libs. Explicitly linking them in flecs' CMake fixed the issue.

CI passed on my fork (unsurprisingly).


SanderMertens:
Thanks for the PR!

PR #764, opened 2022-07-09
Improve documentation and readme

Metadorius:
- Fixed typos, trailing whitespaces, styling inconsistencies all over the documentation, readme and even some comments (in case of "macro's") specifically;
- Emphasized Flecs' unique features more by moving the hierarchies and relations up the key features list in README and adding another sentence highlighting the core design to brief description;
- Brought back the building and usage mini-section from manual (it wasn't intuitive for it to be there, usually such info is placed on project README);
- Removed the addons section from manual as it a) overlaps with the README one, b) seems to be way outdated from a quick glance on blame and the codebase.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,9 +6,11 @@
 [![Try online](https://img.shields.io/badge/try-online-brightgreen)](https://godbolt.org/z/bs11T3)
 [![Documentation](https://img.shields.io/badge/docs-docsforge-blue)](http://flecs.docsforge.com/)
 
-Flecs is a fast and lightweight Entity Component System that lets you build games and simulations with millions of entities ([join the Discord!](https://discord.gg/BEzP5Rgrrp)). Here are some of the framework's highlights:
+Flecs is a fast and lightweight Entity Component System pattern implementation in C that lets you build games and simulations with millions of entities ([join the Discord!](https://discord.gg/BEzP5Rgrrp)). At it's core it uses a [refined approach to building an ECS](https://ajmmertens.medium.com/why-vanilla-ecs-is-not-enough-d7ed4e3bebe5) and as a consequence it is way more flexible than other implementations. Here are some of the framework's highlights:
 
-- Fast zero dependency [C99 API](https://flecs.docsforge.com/master/api-c/)
+- First open source ECS with full support for [Entity Relationships](https://ajmmertens.medium.com/building-games-in-ecs-with-entity-relationships-657275ba2c6c)!
+- Fast native support for [hierarchies](https://flecs.docsforge.com/master/relations-manual/#the-childof-relation) and [prefabs](https://flecs.docsforge.com/master/relations-manual/#the-isa-relation)
+- Fast and [portable](#language-bindings) zero dependency [C99 API](https://flecs.docsforge.com/master/api-c/)
 - Modern type-safe [C++11 API](https://flecs.docsforge.com/master/api-cpp/) that doesn't use STL containers
 - Minimal ECS core with optional [addons](#addons)
 - Entire codebase builds in less than 5 seconds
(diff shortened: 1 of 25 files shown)


README.md
```diff
@@ -6,41 +6,40 @@
 [![Try online](https://img.shields.io/badge/try-online-brightgreen)](https://godbolt.org/z/bs11T3)
 [![Documentation](https://img.shields.io/badge/docs-docsforge-blue)](http://flecs.docsforge.com/)
 
-Flecs is a fast and lightweight Entity Component System that lets you build games and simulations with millions of entities ([join the Discord!](https://discord.gg/BEzP5Rgrrp)). Here are some of the framework's highlights:
+Flecs is a fast and lightweight Entity Component System pattern implementation in C that lets you build games and simulations with millions of entities ([join the Discord!](https://discord.gg/BEzP5Rgrrp)). At it's core it uses a [refined approach to building an ECS](https://ajmmertens.medium.com/why-vanilla-ecs-is-not-enough-d7ed4e3bebe5) and as a consequence it is way more flexible than other implementations. Here are some of the framework's highlights:
```
SanderMertens, line 9: I would like to stick to the old description here for these reasons:
- `pattern implementation in C` - it's both C and C++
- `Entity Component System pattern implementation` - `pattern` doesn't add much here
- `and as a consequence it is way more flexible than other implementations` - probably true, and I've seen other ECS implementations have statements like this in their README. I think it's good form though to let the library speak for itself, instead of dunking (a bit) on others.
- `refined approach` - kind of the same as the previous comment, I don't want to imply that flecs is "better" than others, people can decide that for themselves :)

README.md
```diff
@@ -56,6 +55,13 @@ For more info on ECS, check the [ECS FAQ](https://github.com/SanderMertens/ecs-f
 - [Query Manual](docs/Queries.md) ([docsforge](https://flecs.docsforge.com/master/query-manual/))
 - [Relations Manual](docs/Relations.md) ([docsforge](https://flecs.docsforge.com/master/relations-manual/))
 
+## How to use and build
+The easiest way to add Flecs to a project is to add [flecs.c](https://raw.githubusercontent.com/SanderMertens/flecs/master/flecs.c) and [flecs.h](https://raw.githubusercontent.com/SanderMertens/flecs/master/flecs.h) to your source code. These files can be added to both C and C++ projects (the C++ API is embedded in `flecs.h`).
```
SanderMertens, line 59: Good idea to add this back to the README. I would just add this line, since this comes up a fair amount in the #help channel:

> In C++ projects, make sure to compile the flecs.c file as C code.

README.md
```diff
@@ -56,6 +55,13 @@ For more info on ECS, check the [ECS FAQ](https://github.com/SanderMertens/ecs-f
...
+
+Alternatively you can also build Flecs as a standalone library by using the [bake](https://github.com/SanderMertens/bake) (cmake, meson, bazel are also supported).
```
SanderMertens, line 61: Would change this to:

> Flecs can also be built as a standalone library by using cmake, meson, bazel or [bake](https://github.com/SanderMertens/bake).

I don't want to over-emphasize bake since most people will likely not use it, and I don't want to create the impression that you need bake to use flecs.

README.md
```diff
@@ -56,6 +55,13 @@ For more info on ECS, check the [ECS FAQ](https://github.com/SanderMertens/ecs-f
 - [Query Manual](docs/Queries.md) ([docsforge](https://flecs.docsforge.com/master/query-manual/))
 - [Relations Manual](docs/Relations.md) ([docsforge](https://flecs.docsforge.com/master/relations-manual/))
 
+## How to use and build
```
SanderMertens, line 58: Would change this to `Usage`

README.md
```diff
@@ -56,6 +55,13 @@ For more info on ECS, check the [ECS FAQ](https://github.com/SanderMertens/ecs-f
...
+The easiest way to add Flecs to a project is to add [flecs.c](https://raw.githubusercontent.com/SanderMertens/flecs/master/flecs.c) and [flecs.h](https://raw.githubusercontent.com/SanderMertens/flecs/master/flecs.h) to your source code. These files can be added to both C and C++ projects (the C++ API is embedded in `flecs.h`).
+
+Alternatively you can also build Flecs as a standalone library by using the [bake](https://github.com/SanderMertens/bake) (cmake, meson, bazel are also supported).
+
+If you want to customize your builds, refer to the [Addons](#addons) section.
```
SanderMertens, line 63: Would change this to:

> See the [Addons](#addons) section for learning more about customizing a build.

README.md
```diff
@@ -6,41 +6,40 @@
 [![Try online](https://img.shields.io/badge/try-online-brightgreen)](https://godbolt.org/z/bs11T3)
 [![Documentation](https://img.shields.io/badge/docs-docsforge-blue)](http://flecs.docsforge.com/)
 
-Flecs is a fast and lightweight Entity Component System that lets you build games and simulations with millions of entities ([join the Discord!](https://discord.gg/BEzP5Rgrrp)). Here are some of the framework's highlights:
+Flecs is a fast and lightweight Entity Component System pattern implementation in C that lets you build games and simulations with millions of entities ([join the Discord!](https://discord.gg/BEzP5Rgrrp)). At it's core it uses a [refined approach to building an ECS](https://ajmmertens.medium.com/why-vanilla-ecs-is-not-enough-d7ed4e3bebe5) and as a consequence it is way more flexible than other implementations. Here are some of the framework's highlights:
 
-- Fast zero dependency [C99 API](https://flecs.docsforge.com/master/api-c/)
+- First open source ECS with full support for [Entity Relationships](https://ajmmertens.medium.com/building-games-in-ecs-with-entity-relationships-657275ba2c6c)!
```
SanderMertens, line 11: I see what you're doing here with putting these items further up the list, and I think it's a good idea. I would still put the C/C++ items on top though, since that is the first and most important criteria for anyone using the library.

SanderMertens (changes requested):

README.md
Metadorius: ~~Let's just say I am doing those changes as an outsider so you're not guilty of dunking on others :P~~

Yeah I see what you mean, agree with most of what you said but I would still want to somehow emphasize that it uses an advancement from a traditional ECS pattern you described in the blog, so I would still want to leave that additional sentence, maybe in some a bit more humble form?

How should we call that approach, since it's exactly an an advancement from the core idea of ECS?

README.md
Metadorius: I thought a bit of self-promotion wouldn't hurt here, ~~and you can be free of guilt as well by not putting it here yourself xP~~, but yeah, no problem.

SanderMertens (approved):

PR #789, opened 2022-08-22
Fix Windows headers case to make MinGW compiler happy

lockie:
Hello! This PR makes it possible to compile Flecs using MinGW (since MinGW's own compiler, GCC, is case-sensitive regarding header names).


SanderMertens:
Thanks for the PR!

lockie:
You're welcome :blush:

PR #790, opened 2022-08-23
Fix pipe rendering in Queries.md

MewSoul:
## Description

The `Operator Overview` table in `Queries.md` doesn't render properly because of `|`.
When writing a pipe in a code block, it needs to be espaced with `\` as per the [GFM spec](https://github.github.com/gfm/#example-200).

## Changelog
- Escaped `|` in the `Operator Overview` table to render them properly 

## Results

Before
image

After
image


SanderMertens:
Thanks for the PR!

PR #793, opened 2022-08-28
meta: improve cursor api

randy408:
Allow more types to be set with `ecs_meta_set_(u)int()`
Allow fields to be set to `NULL` with integers


SanderMertens:
Thanks for the PR!

PR #811, opened 2022-09-12
Use stage argument in filter::get_iter method

sro5h:
I think this was an oversight in the latest commit :)


SanderMertens:
You're right! Thanks for the PR :)

Issue #818, opened 2022-09-20
Avoid the boilerplate of .module call inside each module constructor


Hexlord:
**Describe the problem you are trying to solve.**
When creating a C++ module, you have to write `world.module<ModuleTypeName>()` at the beginning, if you forget to do that, the systems will not be children of said module. There are no scenarios where you would not want to scope them like this that are known.

**Describe the solution you'd like**
`world.import<ModuleTypeName>` should automatically scope the systems and components inside the imported module.

SanderMertens:
Fixed!

PR #819, opened 2022-09-20
#818 Scope into the module being imported automatically (20.09 14:06)

Hexlord:
Implements #818 , works fine with excessive `world.module<T>()` call remaining in the constructor, correctly applies `ChildOf(Module)`, saves from having to ever write or remember to write `world.module<T>()` again.

**Cons:** Limits the abilities of the users to decide the scope for components and systems in their modules, so a better solution might be a different function called `.import_scoped<T>()`.


SanderMertens:
Thanks for the PR! I fixed it here: https://github.com/SanderMertens/flecs/commit/4dac95d7c5b7536957dcf5214cbdf139cb538f22 which is almost the same solution, but I had to add an additional line to the C module initialization to make sure interop doesn't break.

PR #826, opened 2022-09-26
musl libc support: use execinfo on Linux only if __GLIBC__ is defined

StaticSaga:
The musl libc does not support backtrace(), but also offers no way of detecting if the code is being compiled with musl headers, hence the only way of detecting the presence of musl is to detect the absence of glibc.
I have not tested this code anywhere but on Linux, but I don't think it can break anything.


SanderMertens:
Looks good! Few small nitpicks:
- it'd be nice if the `#if` before the include would only have to check `HAVE_EXECINFO`
- I use the `_H` postfix for header guards, so `HAVE_EXECINFO` would be better
- Use `ECS_TARGET_LINUX` instead of `__linux__` for consistency

So:

```c
#if defined(ECS_TARGET_LINUX) && !defined(__GLIBC__)
#define HAVE_EXECINFO 0
#elif !defined(ECS_TARGET_WINDOWS) !defined(ECS_TARGET_EM) && !defined(ECS_TARGET_ANDROID)
#define HAVE_EXECINFO 1
#endif

#ifdef HAVE_EXECINFO
#include <execinfo.h>
```

instead of

```c
#if defined(__linux__) && !defined(__GLIBC__)
#define HAVE_EXECINFO_H 0
#else
#define HAVE_EXECINFO_H 1
#endif

#if !defined(ECS_TARGET_WINDOWS) && !defined(ECS_TARGET_EM) && !defined(ECS_TARGET_ANDROID) && HAVE_EXECINFO_H
#include <execinfo.h>
```

Thanks for the PR!

PR #835, opened 2022-10-12
meta_c: expand struct definition before stringifying

randy408:
This should make it possible to use macros inside struct definitions:
```c
struct foo
{
   float bar[BAZ]; //This should work with this patch
};
```
Needs a test case and adjusting the rest of the type macros

### DIFF
--- a/include/flecs/addons/meta_c.h
+++ b/include/flecs/addons/meta_c.h
@@ -44,9 +44,11 @@ extern "C" {
     ecs_meta_from_desc(world, ecs_id(name),\
         FLECS__##name##_kind, FLECS__##name##_desc)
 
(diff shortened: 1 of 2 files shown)


SanderMertens:
I'm closing this PR as it has too many unpredictable side effects, such as platforms that use macro's as type aliases (such as `bool` being defined as `_Bool`). A better solution (as discussed here https://discord.com/channels/633826290415435777/704505897539338251/1030640724376571924) would be to pass both the expanded and non-expanded string to the parser and have the parser use type definitions from the unexpanded string and remaining unresolved identifiers from the expanded string.

That is a much bigger change however, and could take a while to implement (though I will accept a PR that does it =)

PR #847, opened 2022-10-28
Add missing semicolon

nimble0:


SanderMertens:
Nice catch, how did that even compile!

PR #848, opened 2022-10-30
Update BUILD to support .inl files in Bazel builds

benjamind:
I couldn't get building with Bazel to work due to `.inl` file not found errors. This seems like the necessary fix though I'm only just starting to get to grips with Bazel so please let me know if there's an alternative.


SanderMertens:
Thanks for the PR!

PR #854, opened 2022-11-04
Create the same folder structure in Visual Studio as on disk

Tisten:
Made the CMake project generate the same folder structure in Visual Studio (and probably other IDEs) as is used on disk.
This makes it easier to reason about which file is which and to find the files in the solution explorer


SanderMertens:
Thanks for the PR!

PR #866, opened 2022-11-18
Fix missing include

logankaser:
Fix missing include. I use a very strict build environment, which is probably why this was caught.
I suspect most users would not even be able to observe this.


SanderMertens:
This shouldn't be necessary. `misc.c` includes `private_api.h` which includes `private_types.h`, which includes `stdio.h`. What error do you get without the include?

PR #868, opened 2022-11-27
`flecs::entity_t` vs `flecs::entity`

Eshnek:
On my machine using MSVC I had to change the parameter types to `flecs::entity_t` here in order for it to compile.
I am not certain if the C++ API functor signature should be changed, if these documentations should be updated, or if I set it up incorrectly.

```
error C2664: 'Base &flecs::query_builder_i<Base>::order_by(flecs::entity_t,int (__cdecl *)(flecs::entity_t,const void *,flecs::entity_t,const void *))': cannot convert argument 2 from 'M::Pipeline::{ctor}::<lambda_1>' to 'int (__cdecl *)(flecs::entity_t,const void 
*,flecs::entity_t,const void *)'
```

### DIFF
--- a/docs/Queries.md
+++ b/docs/Queries.md
@@ -2493,7 +2493,7 @@ Queries may specify zero for component id to sort on entity ids:
 
 ```cpp
 auto q = world.query_builder<Position>()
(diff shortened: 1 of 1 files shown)


SanderMertens:
Ah yep, that's an issue in the docs. Thanks for the PR!

PR #871, opened 2022-11-29
Use `ecs_ftime_t`

Eshnek:
I have changed `ecs_float_t` to `double` on my configuration. Spares me from
```cpp
cxx/lib/flecs/src/addons/system/system.c:124:33: warning: implicit conversion from ‘float’ to ‘double’ to match other operand of binary expression [-Wdouble-promotion]
  124 |         system_data->time_spent += (float)ecs_time_measure(&time_start);
      |                                 ^~
```
(GCC)

### DIFF
--- a/src/addons/system/system.c
+++ b/src/addons/system/system.c
@@ -121,7 +121,7 @@ ecs_entity_t ecs_run_intern(
     }
 
     if (measure_time) {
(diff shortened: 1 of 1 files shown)


SanderMertens:
Nice catch!

PR #877, opened 2022-12-07
Intentially non-const?

Eshnek:
If I can do `entity.has ( flecs::Disabled )` from a const entity, shouldn't these also be const?


SanderMertens:
Yep they should be! Thx for the PR

PR #884, opened 2022-12-15
Constness

Eshnek:
I think it makes sense for these to be const?

If so, there are some other members that should likely be adjusted as well.


SanderMertens:
Yep! They can be const 😊

Eshnek:
@SanderMertens I added more `const`

SanderMertens:
Awesome, thanks for the PR!

PR #902, opened 2023-01-04
Remove unnecessary backslash in ECS_MODULE macro

copygirl:
I ran into this by chance when trying to parse `#define`s from the header file directly. Seems to be the only place an extra `\` character is at the end of a macro, so likely unintentional.


SanderMertens:
Thanks for the PR!

PR #903, opened 2023-01-04
Match comment on is_monitor with other multi-line comments

copygirl:
I found `is_monitor` is the only place that, for some reason, uses `**<` on subsequent lines of a multi-line comment. I changed it to match the others.


SanderMertens:
Thanks for the PR!

PR #904, opened 2023-01-06
Documentation comments and style suggestions

copygirl:
These are a couple of things that I came across while digging through comments.

- ~~Removing trailing whitespace seems like a good idea. In some cases in the code is used to indent a line with the next one that has code or comments on it, but it's pretty inconsistent. There are also plenty of places where additional spacing was left in accidentally. (This is done for the entire repository, not just source files.)~~ (Removed on request.)
- Some comments were documentation comments where it did not make sense. Maybe they could be converted to group / header comments for doxygen - unsure how that works - but I just turned them into regular comments for now.
- In a couple of places have comments where it seems like it would make sense for them to be documentation comments instead.
- I took the liberty to also make some minor cosmetic changes to align comments where it made sense or moved them around a little to match the surrounding comments better.
- I did not touch the `flecs.c` / `flecs.h` in the root folder, I figure the changes will make it in when you update them the next time around, so I'll leave that up to you.
- When it comes to comments, I focused on the C header files, since that'll end up in the documentation. I found one likely incorrect comment in a `.hpp` file through search, but I haven't gone through any of the C++ files.
- Finally, I included a commit for an `.editorconfig` file. It's editor-agnostic, supported by some IDEs, to inform them which settings to use. One of the settings should have trailing whitespace trimmed automatically on saving, for example.

I split the suggested changes into multiple commits for easier review, potential changes to be made or entire commits to be left out - I wasn't sure about the last one for example. You may want to squash these before merging.

I can make additional changes or remove commits on request.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,9 @@
+root = true
+
+[*]
+charset = utf-8
+indent_style = space
+indent_size  = 4
+end_of_line  = lf
+insert_final_newline = true
+trim_trailing_whitespace = false
--- a/include/flecs.h
+++ b/include/flecs.h
@@ -459,7 +459,7 @@ typedef void (*ecs_poly_dtor_t)(
 /** Iterable mixin.
  * Allows its container to be iterated. */
 typedef struct ecs_iterable_t {
-    ecs_iter_init_action_t init; /* Callback that creates iterator. */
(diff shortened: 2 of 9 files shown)


SanderMertens:
Awesome, thanks for the PR!

PR #906, opened 2023-01-07
Fix clang compilation error

gorilux:
Fix compilation error with the latest emscripten em++:
```
../subprojects/flecs/include/flecs/private/../addons/cpp/utils/enum.hpp:23:30: error: integer value 128 is outside the valid range of values [0, 7] for this enumeration type [-Wenum-constexpr-conversion]
  static constexpr E value = static_cast<E>(Value);
```
Searching around in github I found the solution in the magic_enum project and adapted it.


SanderMertens:
Thanks for the PR!

PR #908, opened 2023-01-12
Add flecs::world::get_ref()

dmlary:
Add flecs::world::get_ref(), and test to verify it is working.


dmlary:
The only failure here was due to a timeout downloading packages.  @SanderMertens, could you kick that one job again?  GitHub isn’t allowing me to do it.

SanderMertens:
Thanks for the PR!

Issue #911, opened 2023-01-16
Proposal: quality-of-life improvements for move-only components


dmlary:
## Overview
Working with move-only components is complex in flecs; mistakes lead to runtime
asserts instead of compile-time errors.  I'm propose a handful of solutions to
simplify the complexity for users of the library.

Most of this issue will be addressing the C++ API, but one of the options
below does describe a broader change at the C level instead.

## Problem
Working with components that are not copyable have some pitfalls in flecs right
now.

* `flecs::entity::get_mut()` inconsistency
    * Calling `get_mut()` inside a read-only system results in a runtime assert
    * Calling `get_mut()` in a read-write system works
    * Calling `get_mut()` outside of a system works
* `flecs::entity::get_ref()` seems to be intended for caching lookups to get to
    a component, but is the only mechanism to get at move-only components
* PreFabs
    * To allow a prefab component to be customized in a child entity, the
      component must be copyable
    * Semantics of copying move-only components for a prefab child differ from
      POD components
        * move-only component is allocating new instance of external resources
          and duplicating data
    * Adding copy to move-only component to support modifiable prefab risks
      resource duplication by allowing `get_mut()` to copy instead of move the
      component
* snapshots
    * there's probably something here too, but I haven't looked at snapshots

All of this boils down to increased cognative load for the developer to
remember when/where it is safe to use `get_mut()`.  Using `get_mut()` in the
wrong scenario results in either a runtime assert, or duplication of external
resources.

## Solutions

Conceptually, move-only components behave significantly different than POD
components in flecs.  At a high level, they should either use separate
functions for modification (partially the case now with `get_ref()`, but not
enforced), or the existing `get_mut()` should work for both POD and move-only
components.

One key point is that to support modification of move-only prefab components,
the component **must** implement a copy constructor.  As a result, flecs cannot
rely just on the presence of the copy constructor to determine if it's safe to
copy a component (expanded on this idea in Option D).

Below are different options that can ease the experience of working with
move-only components for users of the library.  They vary in complexity, but
any one of them will greatly improve working with flecs.

## Option A: `static_assert()` in `flecs::entity::get_mut()` for non-copyable
types

Ok, this is a big blunt hammer that is a breaking change.
Throw a
`static_assert()` for `std::is_copy_constructible` in
`flecs::entity::get_mut()` to force all move-only components to use
`get_ref()`.

Not exactly my preferred approach as it breaks anyone's code that is using 
`get_mut()` in those places where it works.

[...]

PR #915, opened 2023-01-18
ecs_get_mut_id() returns live data where possible

dmlary:
From discussion in issue #911, `get_mut()` will now return live data when the component is present in the entity, or the stage is not deferred.

When the component does not already exist on the entity, and the system is deferred, the component will be created in the command queue and returned.

This commit does not fix the problem of multiple `get_mut()` calls getting different pointers prior to a pipeline sync.


SanderMertens:
It's encouraging that this has so little impact on the tests, it's usually a sign that real applications won't have too many issues with breaking changes either.

A few things to consider:
- Should any special behavior be introduced for multithreaded apps? I'm leaning no, since it would be no more or less safe than using a regular `get` from a system. Future schedule improvements could potentially be introduced to guarantee safety.
- The old code in the `flecs_defer_set` function that invokes the `copy` hook is probably redundant now and can be removed: https://github.com/SanderMertens/flecs/blob/master/src/stage.c#L411
- I'd like to see a few new tests that test the new behavior:
  - When a component exists `get_mut` should be equal to `get`
  - The `EcsOnSet` event of an `ecs_set` should not happen until commands are executed
  - See the `DeferredActions` test suite for how to do this without having to create a system
- Documenting this behavior is needed, but I think we can punt that until after the PR, since documentation in general is a bit light on what exactly happens when operations are deferred.

dmlary:
> Should any special behavior be introduced for multithreaded apps? I'm leaning no, since it would be no more or less safe than using a regular get from a system. Future schedule improvements could potentially be introduced to guarantee safety.

Agreed. At the very least this allows multithreaded components to use synchronization primitives in the component and sync it themselves. Prior to this change, get_mut would attempt to copy. 

> The old code in the flecs_defer_set function that invokes the copy hook is probably redundant now and can be removed:

Will do. 

> No EcsOnSet command is inserted for an ecs_set on an existing component. This is testable by writing the component value after the ecs_set, and making sure it doesn't get overwritten when commands are executed.

I’m a little confused by this one. I’m guessing `ecs_set` (not looking at code right now) uses `ecs_get_mut`. With this change, wouldn’t we still want OnSet to be called, regardless of if the component already existed?  I’m missing something. 

Or is this about deferring the OnSet?

> Documenting this behavior is needed, but I think we can punt that until after the PR, since documentation in general is a bit light on what exactly happens when operations are deferred.

That makes sense deferred behavior isn’t well documented, and this just changes the deferred behavior of `get_mut`

dmlary:
Ok, I think maybe I see what is going on here.  I made my change in `ecs_get_mut_id()`.  All of the `ecs_set*()` functions end up in `flecs_get_mut()`.  Is that where you wanted me to make this change?

dmlary:
> The old code in the flecs_defer_set function that invokes the copy hook is probably redundant now and can be removed:

I'm looking at the code, and reluctant to remove the copy_ctor call at line 411.  I *think* I can see a path where that would be called:
```c
ecs_set_id() /* while deferred, component has a copy constructor, and instance passed in */
flecs_copy_ptr_w_id() /* pass through instance pointer */
flecs_defer_set() /* value != NULL, and component has copy constructor */
```

Is there some condition I've overlooked that would prevent us from reaching 411?  `value != NULL && emplace == false` along that code path.

SanderMertens:
> Ok, I think maybe I see what is going on here. I made my change in ecs_get_mut_id(). All of the ecs_set*() functions end up in flecs_get_mut(). Did you want the change to be in flecs_get_mut() instead of ecs_get_mut_id()?

I don't think it should be in `flecs_get_mut`, because that function is called after the defer check. One option would be to implement it as the first step of the `flecs_defer_set` function.

If the `value` parameter of the `flecs_defer_set` function is `NULL`, it means it's a `get_mut` or `emplace`. What you could do is:

1. Move the code to find the existing component to `flecs_defer_set` (you can just use `ecs_get_id` here)
2. If the component exists and `value` is `NULL`, you can return it 
3. If the component exists and `value` is not `NULL`, copy `value` to the existing component and ensure a `Modified` command is inserted
4. If the component does not exist, use the existing code path to insert a command

I also just thought of an edge that we need to address and write a test for. What if:
- there is an existing component
- a remove command is inserted
- a `get_mut`/`set` is ran afterwards

Because the `get_mut`/`set` come after the `remove`, it should prevent removing the component. It could be that this already works automatically because of command batching (the `remove` would get cancelled out by the `get_mut`/`set`), but we should definitely make sure that this is the case in a test.

dmlary:
I think I'm still missing something.  My change was only intended to apply to `ecs_get_mut()`.

Do you want me to also update the `ecs_set()` codepath to always use the live memory address instead of deferring if the component is present?

src/entity.c
```diff
@@ -2914,16 +2924,32 @@ void flecs_copy_ptr_w_id(
...
+
+    /* lookup the component pointer if the entity exists */
+    if (r->table) {
+        dst = flecs_get_component_ptr(
+                world, r->table, ECS_RECORD_TO_ROW(r->row), id);
+    }
+
+    /* if the component doesn't exist, and we're deferred, copy the new
+     * component to a temporary component on the command queue */
+    bool deferred = flecs_defer_cmd(stage);
```
SanderMertens, line 2939: I think I see why you moved the logic here (to avoid doing a duplicate lookup of `ecs_record_t`), but I'd like the logic for command creation/insertion to stay in `stage.c`. What you could do here instead is lookup the record here, and pass it as an argument to `flecs_defer_set`.

SanderMertens (changes requested):

src/entity.c
dmlary: I don't understand the context of this comment.  Is it in reference to the direct calls to the `flecs_cmd_new()` calls below?

More questions:
* Are you asking for a new argument to be added to `flecs_defer_set()` to pass a record?
*  There's currently no way to call `flecs_defer_set()` without copying or moving the value
    * even with `size == 0`, if there's a copy or move constructor, the size is ignored

Or is something like this more what you're looking for?
```cpp
if (deferred) {
    flecs_defer_set(world, stage, EcsOpAdd, entity, id, 0, NULL, false);
    flecs_defer_set(world, stage, EcsOpModified, entity, id, 0, NULL, false);
}
```

SanderMertens: From Discord (for completeness): the idea is to keep all the defer logic in `flecs_defer_set`, so that the calling code still looks like this:

```c
static
void flecs_copy_ptr_w_id(
    ecs_world_t *world,
    ecs_stage_t *stage,
    ecs_entity_t entity,
    ecs_id_t id,
    size_t size,
    void *ptr)
{
    ecs_record_t *r = flecs_entities_get(world, entity);
    if (flecs_defer_cmd(stage)) {
        flecs_defer_set(world, stage, EcsOpSet, r, entity, id, 
            flecs_utosize(size), ptr, false);
        return;
    }

    // regular undeferred behavior
```

SanderMertens (approved):

PR #928, opened 2023-02-21
Fix the destructor not being called when a column is merged

Thinkofname:
Fixes an issue that happens when the world is destructed (and maybe other cases too).

In the case the column was merged into a non-empty column, the old column's data (after being moved out) was freed but never had the destructor called on it leaking memory with some types.


SanderMertens:
Nice catch! Thanks for the PR and test :)

Issue #929, opened 2023-02-22
Can't lookup an entity by name if it is parented


Tisten:
**Describe the bug**
I was reading https://github.com/SanderMertens/flecs/blob/master/docs/Manual.md#entity-names about how a name can be assigned and "Names can be used to lookup entities". But in my code I couldn't get this to work. And after looking at the tests, and poking at it for a while, I figured out that I can't do `ecs_lookup(world, "entity_name")` on an entity which has been set as a child to another.

**To Reproduce**
Check the PR https://github.com/SanderMertens/flecs/pull/930

**Expected behavior**
Based on the documentation I would expect the name based lookup (i.e only the unique name, not the path) to always work, no matter if the entity has a relationship or not.

SanderMertens:
The reason is this: https://www.flecs.dev/flecs/md_docs_FAQ.html#autotoc_md68

When you add a parent to an entity, you need to use a scoped name to lookup the entity. Try:
```c
ecs_lookup_fullpath(world, "parent.child");
```

SanderMertens:
The reason for this is that name lookups use an index (a hashmap keyed by the name string). This index is created for each subtree of the hierarchy, so that entities in different subtrees can have the same name.

If you need a global identifier that's not specific to a parent, take a look at:
https://www.flecs.dev/flecs/group__paths.html#ga3617d1f76eb10e620665430062425b4c

PR #930, opened 2023-02-22
Created failing test for name based lookups of a child

Tisten:
A test to showcase the issue https://github.com/SanderMertens/flecs/issues/929


SanderMertens:
This is already addressed by tests here: https://github.com/SanderMertens/flecs/blob/master/test/api/src/Hierarchies.c#L382

The reason the test fails is that a scoped name must be passed into the lookup function (e.g. "parent.child").

PR #937, opened 2023-03-05
[docs] Fix conflicting declaration

icolwell:
Noticed this while reading the docs. I assume it's an unintentional copy-paste?

EDIT: I re-read and realized it was probably supposed to be 2 different ways of adding components, not one continuous code example. I added some comments to clarify that instead of removing the `auto` declaration.

### DIFF
--- a/docs/Quickstart.md
+++ b/docs/Quickstart.md
@@ -191,19 +191,16 @@ const Position *p = ecs_get(world, e, Position);
 ecs_remove(world, e, Position);
 ```
 ```cpp
-// Option 1: Create entity and assign component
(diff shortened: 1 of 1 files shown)


SanderMertens:
Nice catch! I think it's better to use the existing `e` from the example, so it becomes:
```cpp
e.set<Position>({10, 20})
  .set<Velocity>({1, 2});
```

That way the comments & structure is the as the C example above it.

icolwell:
Ah, that makes more sense, updated!

SanderMertens:

PR #939, opened 2023-03-08
Provide row in OnSet

Hexlord:
This info is useful in observers


SanderMertens:

PR #943, opened 2023-03-18
ecs_meta_set_int(): fix unsigned bounds check

randy408:
Before this fix if you passed e.g. `-1` to set an U64 it will fail, but all values are valid for an U64. It will still error out for U32/16/8 with a value of `-1` but I think that's fair.


SanderMertens:
Thanks for the PR!

PR #955, opened 2023-04-03
Add a way to get the filter from a rule, swap str method to rule specific one

Thinkofname:
Noticed the `str` method for rules wasn't `const` so went to fix that and noticed in the process that it was calling the filter string method instead of the new one for rules so I've also changed it to that whilst also adding a method to get the filter so the old string result can still be fetched.

This is a breaking change however from the previous behavior though due to the output being very different so it might be better to add a different method to return the rule string?

### DIFF
--- a/flecs.h
+++ b/flecs.h
@@ -27734,9 +27734,12 @@ struct rule_base {
         obj.m_rule = nullptr;
     }
 
-    flecs::string str() {
-        const ecs_filter_t *f = ecs_rule_get_filter(m_rule);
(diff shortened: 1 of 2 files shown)


SanderMertens:
Yup, I would keep the old behavior of the `str()` method so that it behaves the same as `filter::str`.

Thinkofname:
Changed back and added `rule_str` to cover the other method

SanderMertens:

PR #960, opened 2023-04-11
Added example for scene management using pipelines

benjitrosch:


examples/cpp/game_mechanics/scene_management/src/main.cpp
```diff
@@ -0,0 +1,189 @@
...
+struct Position { float x, y; };
+struct Button { std::string text; };
+struct Character { bool alive; };
+struct Health { int amount; };
+
+// Removes all entities who are children of
+// the current scene root and returns it.
+// (NOTE: should use defer_begin() / defer_end())
+flecs::entity reset_scene(flecs::world& ecs) {
+    auto scene = ecs.component<SceneRoot>();
```
SanderMertens, line 26: You could do this here, which is a bit more efficient:

```cpp
ecs.delete_with(flecs::ChildOf, ecs.entity<SceneRoot>());
```

SanderMertens (changes requested):

examples/cpp/game_mechanics/scene_management/src/main.cpp
benjitrosch: Thanks, fixed with 1217d63

SanderMertens (approved):

PR #964, opened 2023-04-20
Add override_second and set_override_second APIs

InfRandomness:
Hey everyone 👋 

I'm opening this PR because I was suggested by a Discord member to create this API as a newcommer PR to make.

I've created the override_second and set_override_second APIs, both equally using their "second" counterpart APIs

I only created the APIs I thought were useful, being unsure if the other APIs such as 
https://github.com/SanderMertens/flecs/blob/master/include/flecs/addons/cpp/mixins/entity/builder.hpp#L375-L384
or
https://github.com/SanderMertens/flecs/blob/master/include/flecs/addons/cpp/mixins/entity/builder.hpp#L386-L395
also need replication.

:)

### DIFF
--- a/include/flecs/addons/cpp/mixins/entity/builder.hpp
+++ b/include/flecs/addons/cpp/mixins/entity/builder.hpp
@@ -408,11 +408,11 @@ struct entity_builder : entity_view {
     /** Mark component for auto-overriding.
      * @see override(flecs::id_t id)
      *
-     * @tparam T The component to mark for overriding.
+     * @tparam Second The component to mark for overriding.
      */
-    template <typename T>
-    Self& override_second() {
-        return this->override_second(_::cpp_type<T>::id(this->m_world));
+    template <typename Second>
+    Self& override_second(flecs::entity_t first) {
+        return this->override(first, _::cpp_type<Second>::id(this->m_world));
     }
 
     /** Mark pair for auto-overriding.
--- a/include/flecs/addons/cpp/mixins/entity/builder.hpp
+++ b/include/flecs/addons/cpp/mixins/entity/builder.hpp
@@ -394,6 +394,35 @@ struct entity_builder : entity_view {
         return this->override<First>(_::cpp_type<Second>::id(this->m_world));
     }
 
+    /** Mark id for auto-overriding.
+     * When an entity inherits from a base entity (using the IsA relationship)
+     * any ids marked for auto-overriding on the base will be overridden
+     * automatically by the entity.
+     *
+     * @param id The id to mark for overriding.
+     */
+    Self& override_second(flecs::id_t id) {
+        return this->add_second(ECS_OVERRIDE | id);
+    }
+
+    /** Mark component for auto-overriding.
+     * @see override(flecs::id_t id)
+     *
+     * @tparam T The component to mark for overriding.
+     */
+    template <typename T>
+    Self& override_second() {
+        return this->override_second(_::cpp_type<T>::id(this->m_world));
+    }
+
+    /** Mark pair for auto-overriding.
+     * @param first The first element of the pair.
+     * @param second The second element of the pair.
+     */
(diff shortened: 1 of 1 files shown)


include/flecs/addons/cpp/mixins/entity/builder.hpp
```diff
@@ -394,6 +394,35 @@ struct entity_builder : entity_view {
...
 
+    /** Mark id for auto-overriding.
+     * When an entity inherits from a base entity (using the IsA relationship)
+     * any ids marked for auto-overriding on the base will be overridden
+     * automatically by the entity.
+     *
+     * @param id The id to mark for overriding.
+     */
+    Self& override_second(flecs::id_t id) {
+        return this->add_second(ECS_OVERRIDE | id);
```
SanderMertens, line 405: I don't think this works, the `add_second` method requires a template argument:
```cpp
    template<typename Second>
    Self& add_second(flecs::entity_t first) {
        return this->add(first, _::cpp_type<Second>::id(this->m_world));
    }
```

include/flecs/addons/cpp/mixins/entity/builder.hpp
```diff
@@ -394,6 +394,35 @@ struct entity_builder : entity_view {
...
+    template <typename Second>
+    Self& override_second(flecs::entity_t first) {
+        return this->override(first, _::cpp_type<Second>::id(this->m_world));
+    }
+
+    /** Mark pair for auto-overriding.
+     * @param first The first element of the pair.
+     * @param second The second element of the pair.
+     */
+    Self& override_second(flecs::entity_t first, flecs::entity_t second) {
```
SanderMertens, line 422: I don't think this method is necessary. The `*_second` functions are necessary when the first element of a pair is an id, and the second element of a pair is a type. In this case both elements are ids, which makes this function a duplicate of:

```cpp
    Self& override(flecs::entity_t first, flecs::entity_t second) {
        return this->override(ecs_pair(first, second));
    }
```

include/flecs/addons/cpp/mixins/entity/builder.hpp
```diff
@@ -501,6 +530,17 @@ struct entity_builder : entity_view {
         return to_base();  
     }
 
+     /** Set component, mark component for auto-overriding.
+      * @see override(flexs::id_t id)
+      *
+      * @tparam T the component to set and for which to add the OVERRIDE flag
+      */
+     template<typename T>
+     Self& set_override_second(const T& val) {
```
SanderMertens, line 539: Methods with `_second` operate on pairs, which this method doesn't. You're missing a `flecs::entity_t first` argument. Additionally, to make sure `set` works in all cases (by value, by ref, move ref etc) a number of overloads have to be implemented, see `set_override` for examples.

SanderMertens (changes requested):

InfRandomness:
I will correct these today (hopefully), 👍

include/flecs/addons/cpp/mixins/entity/builder.hpp
InfRandomness: I'm not sure what I should be passing over to add_second 🤔

include/flecs/addons/cpp/mixins/entity/builder.hpp
InfRandomness: are all of those overloads needed?

include/flecs/addons/cpp/mixins/entity/builder.hpp
MewSoul: Probably something like that:
```cpp
    template <typename Second>
    Self& override_second(flecs::id_t id) {
        return this->add_second<Second>(ECS_OVERRIDE | id);
    }
```

SanderMertens:
Closing PR for inactivity.

PR #974, opened 2023-04-29
Remove broken link. I got rickrolled

elnawe:
I'm proposing to remove this link from here since it's broken or changed destination.


SanderMertens:
You're not supposed to say that out loud.

elnawe:
Oops :^P

PR #978, opened 2023-05-02
Emulate posix atomics for systems that do not have gnu built-ins

aganm:
Made for TCC but that should also work for every system/compiler without the gnu builtins.


aganm:
It's built-ins or builtins? I may have used both terms in the same comment xD

SanderMertens:
LGTM!

PR #983, opened 2023-05-09
meson: link ws2_32 for mingw builds

randy408:
With MSVC `#pragma comment(lib, "ws2_32.lib")` is enough, but mingw doesn't support that.

Build error initially reported here https://github.com/flecs-hub/flecs-lua/issues/12

@MasterDrake can you confirm the fix?

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -8,7 +8,15 @@ endif
 
 flecs_inc = include_directories('include')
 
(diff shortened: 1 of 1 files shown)


SanderMertens:
Thanks for the PR! Looks like a low risk change to me.

PR #984, opened 2023-05-09
Remove sep initialization in ecs_new_from_path_w_sep

copygirl:
This appears to be unnecessary because `ecs_add_path_w_sep` already does the same thing.


SanderMertens:
LGTM!

PR #992, opened 2023-06-08
Remove dead code: datastructures/qsort.{c,h}

jbarthelmes:
It looks like it was used in the rules addon but I found no current usage.

You will need to amalgamate yourself.


SanderMertens:
Nice catch!

PR #999, opened 2023-06-30
Don't return malformed JSON from REST API if serialization failed

ZeroErrors:
Previously when generating a reply for the entity and world REST API endpoints, the code ignored any errors, this meant it would send partial JSON data to the client with a 200 OK HTTP response.

This change instead checks if serialization was successful and if not will clear the body of the response and return a 500 Internal server error instead.


SanderMertens:
Thanks for the PR!

PR #1001, opened 2023-07-08
Remove entities support from ecs_field functions

ZeroErrors:
## Changelog
- Removes support for index 0 from:
  - `ecs_field_w_size`
  - `ecs_field_size`
- Ensure all the following functions assert that `index >= 1`
  - `ecs_field_w_size`
  - `ecs_field_is_readonly`
  - `ecs_field_is_writeonly`
  - `ecs_field_is_set`
  - `ecs_field_is_self`
  - `ecs_field_id`
  - `ecs_field_column_index`
  - `ecs_field_src`
  - `ecs_field_size`
- Added `flecs::column<const flecs::entity_t> flecs::iter::entities()` to allow access to the iterators `entities` with the C++ API.
- Added `Query_iter_entities` to test the `flecs::iter::entities()` function.

## Details
I was attempting to write a filter like the following snippet and while doing so discovered that the `flecs::iter` was asserting when trying to access the field for the entities.
```cpp
std::vector<flecs::entity_t> allEntities;
ecs.filter_builder<>()
  .with<SomeTag>()
  .build()
  .iter([&allEntities](flecs::iter& it)
    {
      auto entities = it.field<const flecs::entity_t>(0);
      allEntities.insert(allEntities.end(), &entities[0], &entities[it.count()]);
    });

// Sort the entity ids to ensure the order is stable
std::sort(allEntities.begin(), allEntities.end());

for (auto eId : allEntities)
{
  // ...
}
```
Note that there is a way to work around this by accessing the underlying `ecs_iter_t*`
```cpp
auto iter = it.c_ptr();
allEntities.insert(allEntities.end(), &iter->entities[0], &iter->entities[it.count()]);
```

<br/>

The specific assert I was seeing was the id retrieved from `ecs_field_id` didn't match the id from `_::cpp_type<T>::id(m_iter->world)` causing this assert to fail https://github.com/SanderMertens/flecs/blob/798f7353cdb7659c9795ef458d4b7de3ef45bfd9/include/flecs/addons/cpp/iter.hpp#L421-L426

When looking into this I also found that `ecs_field_id` was doing out-of-bounds access since it was doing `index - 1` and the assert was allowing `index >= 0` which would cause it to do `return it->ids[-1];`
https://github.com/SanderMertens/flecs/blob/798f7353cdb7659c9795ef458d4b7de3ef45bfd9/src/iter.c#L450-L457

From there I decided to check the other `ecs_field_*` functions to make sure they acted correctly with `index == 0` and there were a few more that needed changing to handle this case correctly.

### DIFF
--- a/flecs.c
+++ b/flecs.c
@@ -59455,68 +59455,53 @@ bool flecs_iter_next_instanced(
 void* ecs_field_w_size(
     const ecs_iter_t *it,
     size_t size,
-    int32_t term)
+    int32_t index)
 {
     ecs_check(it->flags & EcsIterIsValid, ECS_INVALID_PARAMETER, NULL);
-    ecs_check(it->ptrs != NULL, ECS_INVALID_PARAMETER, NULL);
-    ecs_check(!size || ecs_field_size(it, term) == size || 
-        (!ecs_field_size(it, term) && (!it->ptrs[term - 1])), 
+    ecs_assert(index >= 1, ECS_INVALID_PARAMETER, NULL);
+
+    ecs_check(!size || ecs_field_size(it, index) == size ||
+        (!ecs_field_size(it, index) && (!it->ptrs[index - 1])),
             ECS_INVALID_PARAMETER, NULL);
     (void)size;
 
-    if (!term) {
-        return it->entities;
-    }
-
-    return it->ptrs[term - 1];
+    return it->ptrs[index - 1];
 error:
     return NULL;
 }
 
 bool ecs_field_is_readonly(
     const ecs_iter_t *it,
-    int32_t term_index)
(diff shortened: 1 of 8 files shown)


SanderMertens:
Hm, I agree that this improves consistency of the code, but it adds branches to support a feature that's very niche. I actually want to remove exposing the entity id array as field 0, it's a leftover from a very old version of Flecs, and not really necessary anymore.

ZeroErrors:
No problem, I was wondering if it would be better to remove support for it instead. I'll change the PR to move in that direction instead.

SanderMertens:
LGTM, thanks for the PR!

PR #1002, opened 2023-07-14
Add world::run_pipeline<T>(ecs_ftime_t) overload

ZeroErrors:
## Changelog
- Add `world::run_pipeline<T>(ecs_ftime_t)` overload

## Details
Was working with custom pipelines and noticed that this convenience overload was missing.

With this change instead of writing
```cpp
ecs.run_pipeline(ecs.id<CustomPipeline>().entity());
```
You can now just do
```cpp
ecs.run_pipeline<CustomPipeline>();
```


SanderMertens:
LGTM, thanks for the PR!

PR #1004, opened 2023-07-16
Add world::target

ZeroErrors:
## Changelog
- Added `world::target(int32_t index)` and `world::target(flecs::entity_t relationship, int32_t index)`

## Details
There was a [question asked on Discord](https://discord.com/channels/633826290415435777/1129888974475886672/1129888974475886672) about how to  get the target entity after doing something like the following:
```cpp
struct ActivePlayer {};
// ...

ecs.component<ActivePlayer>().add(flecs::Exclusive);

// ...

auto player = world.entity();
ecs.add<ActivePlayer>(player);
```

At the moment you can do this with (or some variation of):
```cpp
auto active_player = ecs.singleton<ActivePlayer>().target<ActivePlayer>();
```

With the change in this PR this now becomes more convenient to do, like so:
```cpp
auto active_player = ecs.target<ActivePlayer>();
```

### DIFF
--- a/flecs.h
+++ b/flecs.h
@@ -19520,6 +19520,16 @@ struct world {
     template<typename T>
     flecs::entity target(flecs::entity_t first, int32_t index = 0) const;
 
+    /** Get target for a given pair from a singleton entity.
+     * This operation returns the target for a given pair. The optional
+     * index can be used to iterate through targets, in case the entity has
+     * multiple instances for the same relationship.
+     *
+     * @param first The first element of the pair for which to retrieve the target.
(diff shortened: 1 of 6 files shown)


SanderMertens:
Looks good! Can you add tests for this to the C++ Singleton test suite?

ZeroErrors:
Added the test and also added another overload without the template argument so you don't need an actual type to use it.
```cpp
auto tag = ecs.entity();
ecs.set(tag, entity);

auto entity = ecs.target(tag);
```

SanderMertens:
Nice, thanks for the PR!

PR #1005, opened 2023-07-18
Remove unneeded if and main_thread param from flecs_run_pipeline_ops

ZeroErrors:
I was looking at changes from https://github.com/SanderMertens/flecs/pull/1003 again and realised there was some redundant code that could be cleaned up.

## Changelog
- Removed unneeded `if (main_thread || op->multi_threaded)` from `flecs_run_pipeline_ops`
  - This isn't needed since `flecs_run_pipeline` only signals workers when `op->multi_threaded` is set.
  - I turned this into an assert instead just to make sure it doesn't happen.
- Removed `bool main_thread` parameter from `flecs_run_pipeline_ops` since it is not needed.

### DIFF
--- a/flecs.c
+++ b/flecs.c
@@ -16886,8 +16886,7 @@ int32_t flecs_run_pipeline_ops(
     ecs_stage_t* stage,
     int32_t stage_index,
     int32_t stage_count,
-    ecs_ftime_t delta_time,
-    bool main_thread);
+    ecs_ftime_t delta_time);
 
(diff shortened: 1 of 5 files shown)


SanderMertens:
Hm, how does this work in multithreaded worlds with systems that aren't multithreaded? 🤔 Those should only be ran on the main thread.

SanderMertens:
Ah, I think I see how this works, the worker threads aren't signaled when the active part of the schedule isn't multithreaded- cool! LGTM 🚀

PR #1007, opened 2023-07-19
Add a `ecs_singleton_set_ptr`, to mirror `ecs_set_ptr`

darkuranium:
A trivial change that to fix a source of annoyances in some places.


SanderMertens:
LGTM!

PR #1012, opened 2023-07-28
Add better cursor support in stack allocators

johnse-hypixel:
This PR fixes a bug in the stack allocator that can cause memory corruption in scenarios where multiple (nested) iterators are iterated. 

=== Original message ===
This PR is to fix some tests that have errors hidden by a large bug. These test cases were not properly finishing iteration, nor calling `ecs_iter_fini()`. Furthermore, there are cases of overlapping lifetimes of objects allocated on the stack allocator.

The test failures in subsequent CI runs shows the errors found be adding consistency checks to the stack allocator. Then as the new tracking of freed data improved, and fixes in some test cases, the CI errors were eliminated over time.

The existing test cases exercised the various lifetime issues that the new tracking is designed to support. Behavior of the allocator is:

- new allocations only occur at top of stack even if there are freed spaces further down
- memory is only returned to the stack when the top-of-stack cursor is restored.
- If we have four iterators on the stack: **A** {B} {C} **D** (B&C marked free):
  - Freeing D will return B, C, and D and leave A as the top of stack (A becomes the tailCursor)
  - Allocating E will result in **A, E**

<s>The first commit intentionally has the fixes commented out and additional checks to confirm that the iterator is finished. This first commit will therefore fail CI tests a total of 7 failures.

The second commit will show these are fixed.</s>


src/iter.c
```diff
@@ -841,6 +841,8 @@ ecs_iter_t ecs_page_iter(
     ecs_check(it->next != NULL, ECS_INVALID_PARAMETER, NULL);
 
     ecs_iter_t result = *it;
+    result.priv.cache.stack_cursor = NULL; /* Don't copy allocator cursor */
```
johnse-hypixel, line 844: Yes. These were the places where I ran into issues with storing pointers. Glad you had the context to know what was expected of the wrapper iterator!

johnse-hypixel (commented):
Look great! Very nice refactor to return the cursor pointer. Much less bookkeeping.

I can't hit the "Approve", but I approve these changes!

PR #1037, opened 2023-09-05
Fix dupe word typos

Rageking8:


SanderMertens:
Wow how did you find all of those 😀 Thanks for the PR!

Rageking8:
> Wow how did you find all of those 😀 Thanks for the PR!

Haha. I found those using a crude regex search across the whole codebase targeting a common typo where one repeats a word twice when only one is intended.

Rageking8:
https://github.com/SanderMertens/flecs/blob/ec237bf0152aa17e5bb8fd1a6fe537cdc13e7fe0/flecs.h#L27544
https://github.com/SanderMertens/flecs/blob/ec237bf0152aa17e5bb8fd1a6fe537cdc13e7fe0/include/flecs/addons/cpp/mixins/filter/impl.hpp#L141

Left out some that I was unsure on, such as the above.

PR #1044, opened 2023-09-17
match function parameter names across declaration and definition in entity view

Indra-db:
match function parameter names across declaration and definition in entity view

removing the s_id == id -> we already know s_id is not 0, and if the first check (!id) passed, it's 0, so s_id will never match id and will always assert. Unless this is intended? If so we could put false down instead.


Indra-db:
will do this in another pull request

PR #1053, opened 2023-10-03
bump cmake_minimum_required version

Sororfortuna:
Minor change to CMakelists.

versions before 3.5 are deprecated in latest CMake versions and causes a warning:
```
CMake Deprecation Warning at flecs/CMakeLists.txt:1 (cmake_minimum_required):Compatibility with CMake < 3.5 will be removed from a future version of
CMake.

Update the VERSION argument <min> value or use a ...<max> suffix to tell
CMake that the project does not need compatibility with older versions.
```


SanderMertens:
Thanks for the PR!

PR #1054, opened 2023-10-03
fix/improve some comments, group add and remove functions

Indra-db:
~~I changed the comment of set_lookup_path dramatically to be aligned with the comment on the C function lookup_path, as it was a lot more descriptive. However, I used the @ parameters more heavily to split it up into sections.~~

after discussion with Sander, he instead suggested to use @see so there's no comment duplication across CPP/C API so @see has been added to set_lookup_path cpp function

I relocated the add enum function to be grouped with the other add functions. Same with Remove. 

The rest is quite minor.

### DIFF
--- a/flecs.h
+++ b/flecs.h
@@ -19784,8 +19784,34 @@ struct world {
     template <typename T>
     flecs::entity set_scope() const;
 
-    /** Set search path.
-     */
+    /**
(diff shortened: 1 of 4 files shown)


SanderMertens:
Thanks for the PR!

PR #1055, opened 2023-10-03
remove redundant get_ref component registration ensure

Indra-db:
for the standard get_ref I'm passing it as a parameter, which should ensure registration too

for the other get_refs, it's already being registered in the function parameter passthrough too.


SanderMertens:
Thanks for the PR!

PR #1056, opened 2023-10-03
set name if alias is nullptr for use

Indra-db:
it's worth considering if there should be a default parameter value for alias, which is set to nullptr.

In case you agree to remove it, I'll update the PR and 
`set ecs_check(entity != 0 || name != NULL, ECS_INVALID_PARAMETER, NULL); `
to
`ecs_check(entity != 0 && name != NULL, ECS_INVALID_PARAMETER, NULL);`
in
`flecs_set_identifier` function

In case we do, then the `if(!name)` check in `use` function is quite redundant and we could optionally remove it since flecs_set_identifier already deals will nullptr name where it just removes the previous alias and sets no new (as far as I understand)

### DIFF
--- a/flecs.h
+++ b/flecs.h
@@ -30535,9 +30535,9 @@ inline void world::use(flecs::entity e, const char *alias) const {
     const char *name = alias;
     if (!name) {
         // If no name is defined, use the entity name without the scope
-        ecs_get_name(m_world, eid);
+        name = ecs_get_name(m_world, eid);
     }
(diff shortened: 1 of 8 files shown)


Indra-db:
optionally if it's nullptr, we could just early out and do nothing? 
It seems a bit "bad" to remove the previous alias. I would classify that as unexpected behaviour.

SanderMertens:
Thanks for the PR!

PR #1058, opened 2023-10-03
change return type of world::id to flecs::id instead of flecs::entity…

Indra-db:
… and add missing world parameter passthrough for cpp_type<T>::id in entity_builder::enable


SanderMertens:
Thanks for the PR!

PR #1061, opened 2023-10-10
Improve assert performance

jbarthelmes:
As discussed on discord, moving the condition outside of `ecs_assert` improves debug performance. Now I also tested on ecs_benchmark and got ~1.4x speedups. 

For further work I need your opinion on 2 things:
1. The same could be done to `ecs_poly_assert`, but `ecs_poly` casts the return value from `ecs_poly_assert`. My suggestion is to do the check separately: [Example in my fork](https://github.com/jbarthelmes/flecs/commit/e00a9c15dbc549b34299d0162d12a353f9793206#diff-227b94484bf9df7691c3995c506646ee41f49e57356d50db1b55e00131767d15). Those 5 were the only callsites 2 weeks ago. Can you live with that?
2. Performance could be further improved (I measured ~1.3x) by removing the calls to `ecs_os_abort` and `assert`. This can be done by making the fail handler noreturn and moving the call to abort inside. However, adding noreturn requires some `#if`s because the C standard changed on how to use it: [cppreference](https://en.cppreference.com/w/c/language/_Noreturn). So I'd add a `#define FLECS_NORETURN` depending on std(/compiler?) version. Do you want that?

[My benchmark results are attached](https://github.com/SanderMertens/flecs/files/12857299/assert-bench.ods).
You'll have to amalgamate yourself.

### DIFF
--- a/include/flecs/addons/log.h
+++ b/include/flecs/addons/log.h
@@ -155,8 +155,7 @@ void ecs_abort_(
     ...);
 
 FLECS_API
-bool ecs_assert_(
-    bool condition,
+void ecs_assert_log_(
     int32_t error_code,
     const char *condition_str,
     const char *file,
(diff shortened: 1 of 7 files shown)


jbarthelmes:
done

SanderMertens:
Thanks for the PR!

PR #1064, opened 2023-10-28
Allow for optional pairs in C++ systems

Thinkofname:
Modifies the templates for pairs in order to allow them to handle being used as an optional term in a system. It does this by adding `raw_type_t` to strip both pointers and references for most operations but adding `transcribe_pointer_t` to put them back when needed. This means that most of the code should end up handling this transparently without many changes but I've only tested systems right now.

Only tested with gcc currently as its all I had on me at the time.


SanderMertens:
Looks like it works on all compilers/platforms! Thanks for the PR :)

PR #1074, opened 2023-11-14
make the hello world example optional when building via meson

apache-hb:
prevents the hello world example from being built by default when consumed as a subproject


SanderMertens:
LGTM, thanks for the PR!

PR #1091, opened 2023-12-08
Add template specialization to `function_traits_impl` to handle noexcept free functions. 

MilanDierick:
This seems to be working for the most part. However, a couple things of note: 
- the calling convention need to be specified, which could cause issues with other calling conventions
- this does not seem to work with `noexcept` lambdas

For lambdas, I'm not sure if there's a way to get the argument types or return types, or if we can assume a set of arguments for the lambda for this to work. I tried to think of a way to introduce this: `function_traits_impl<decltype(&Callable::operator())>` but I'm not sure how to go about this, lambdas are tricky. 

We could also just take the small win and add support for free floating `noexcept` functions and leave lambdas as is. However, that could introduce some confusion when it comes to using the API which doesn't seem ideal. 

As for the test, I have no idea if what I did was right, please do confirm I put everything in the right place.

### DIFF
--- a/include/flecs/addons/cpp/utils/function_traits.hpp
+++ b/include/flecs/addons/cpp/utils/function_traits.hpp
@@ -76,6 +76,10 @@ template <typename ClassType, typename ReturnType, typename... Args>
 struct function_traits_impl<ReturnType(ClassType::*)(Args...) const volatile&&>
     : function_traits_defs<ReturnType, Args...> {};
 
+template <typename ReturnType, typename... Args>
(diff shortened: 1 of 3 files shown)


MilanDierick:
I'm scratching my head a bit here; it seems that both macOS and Linux are happy to compile, but window isn't, even though I'm compiling Flecs locally on a Windows machine just fine. Unless you know the cause by just looking at it briefly, I'd say that this isn't worth the effort, a missing `noexcept` on a limited number of callbacks isn't a problem.

PR #1094, opened 2023-12-09
Print doubles as `%f`, not `%d`.

waywardmonkeys:
The `min` and `max` fields are `double`, so we print them with `%f` rather than `%d`.


SanderMertens:
Thanks for the PR!

PR #1101, opened 2023-12-18
Prevent worker threads from being preempted by other tasks on Mac and iOS

staminajim:
When doing some profiling on Mac and iOS to find the cause of stuttery frame rates, I found the threads being created for multithreaded systems were occasionally being preempted for a large chunk of time by both non-application tasks, and from non-render threads in the application itself.

I found this https://github.com/flutter/flutter/issues/65752 and added in the few lines of code found in this WWDC video:
https://developer.apple.com/videos/play/wwdc2018/612/?time=1111 

And now the ECS system threads are no longer pre-empted and the stutters gone!

I'm not sure if this helps on linux or if this needs more thinking about.

It's also kinda hard to prove as the problem relies on the OS deciding to prioritise other work over the flecs threads, but it's working great for me on Mac + iOS

### DIFF
--- a/src/addons/os_api_impl/posix_impl.inl
+++ b/src/addons/os_api_impl/posix_impl.inl
@@ -27,9 +27,19 @@ ecs_os_thread_t posix_thread_new(
 {
     pthread_t *thread = ecs_os_malloc(sizeof(pthread_t));
 
-    if (pthread_create (thread, NULL, callback, arg) != 0) {
+    /* Opt out of Quality of Service and set a high priority 
+     * to prevent preemption by system services. */
+    pthread_attr_t attr;
+    pthread_attr_init (&attr);
+    pthread_attr_setschedpolicy(&attr, SCHED_RR);    
+    struct sched_param param = {.sched_priority = 45};
+    pthread_attr_setschedparam(&attr, &param);
+    
+    if (pthread_create(thread, &attr, callback, arg) != 0) {
         ecs_os_abort();
     }
+
+    pthread_attr_destroy(&attr);
 
     return (ecs_os_thread_t)(uintptr_t)thread;
 }


SanderMertens:
Trying to decide on what to do with this PR. One the one hand this seems like a simple fix, and if it's improving performance then that's a good argument in favor.

On the other hand though (like you mentioned) it is difficult to verify the impact on different platforms, and could result in unexpected behavior for existing applications (like flecs apps consuming too many resources on systems with multiple applications, or permission issues with setting priority). The magic 45 value is also not great, as you'd have no way to customize it.

I think I'm currently leaning towards not merging this, and adding it as a recommendation in one of the manuals. This behavior can be customized easily by overriding the `thread_new_` callback on the OS API.

staminajim:
I wonder if we could get a second performance opinion if there’s any other Mac / iOS / Linux devs on discord?

As far as I know it’s difficult to bog down any of those OSes without going into realtime priority (ie for audio thread callbacks).

Not fussed if it just becomes a note in the manual. I’ll also do some more research when I get back to my computer next week since I really just followed Apple’s example code

ZeroErrors:
Just adding my 2 cents. Personally I feel that setting the thread priority higher than default in a library would generally be bad practice because it makes any software you build with it a "bad citizen" by requesting priority over other applications instead of letting the OS scheduler do its thing. Only a handful of special cases where you actually need real-time or low-latency should ever need to change thread priorities and usually it only yields a performance improvement if there are a bunch of background processes eating up scheduler time.

However, having the ability to change the thread priority for worker threads could be useful for some people and given flecs use cases its more likely for people using flecs to have one of the use cases where using a higher priority would be desired. Like Sander suggested replacing the `thread_new_` in `ecs_os_api_t` already provides a way to do this.

Perhaps flecs could provide a util function to help people do this instead of changing the default, or at least document how to do it in the manual.

SanderMertens:
Closing this- it's useful, but something that is better suited for a custom OS API implementation vs. making it the default.

PR #1103, opened 2023-12-20
added ability to build flecs static lib with exported symbols in cmake

mctb32:
This is an opt-in feature, so the default behavior stays the same and no external projects should break. It is useful when someone is building a DLL and wants to link statically to flecs, but expose flecs api from their DLL. 

To use it you can set the variable in your script prior to importing flecs:

set(FLECS_STATIC_EXPORTS ON CACHE BOOL "")
add_subdirectory(flecs)

The DLL target could then be defined as follows:

add_library(my_dll_target SHARED my_source.c)
target_link_libraries(my_dll_target PRIVATE flecs::flecs_static)
target_include_directories(my_dll_target INTERFACE flecs/include)
target_compile_definitions(my_dll_target INTERFACE flecs_IMPORTS)


mctb32:
I'm closing the pull request. It turns out exporting symbols in a static lib is not practical as when building the dll msvc linker strips unused symbols and you'd need to prepare a custom .def file listing all functions.

PR #1107, opened 2023-12-29
Fix relationship example by specifying inout.

garrett-is-a-swann:
Fixes an example that is currently broken due to a change where .with<T> requests no data by default.

Another fix could have been to use .term<T> instead, but it's useful to see .with<T> and access modifiers in action.

### DIFF
--- a/examples/cpp/relationships/union/src/main.cpp
+++ b/examples/cpp/relationships/union/src/main.cpp
@@ -35,10 +35,12 @@ int main(int argc, char *argv[]) {
     ecs.component<Direction>().add(flecs::Union);
 
     // Create a query that subscribes for all entities that have a Direction
-    // and that are walking
+    // and that are walking.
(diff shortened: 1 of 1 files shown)


SanderMertens:
Ah yeah, that example is broken because of a breaking change a while back. Since the code doesn't actually modify the entity arrays (which should not be allowed, should add a check for that), it should be modified to this:

```cpp
    flecs::query<> q = ecs.query_builder()
        .with(Walking).in()
        .with<Direction>(flecs::Wildcard).in()
        .build();

// ...

        auto movement = it.field<const flecs::entity_t>(1);
        auto direction = it.field<const flecs::entity_t>(2);
```

garrett-is-a-swann:
Oops, duh. Had it backwards. Will make those changes.

SanderMertens:
LGTM, thanks for the PR!

PR #1108, opened 2023-12-29
Update doxygen

ZeroErrors:
See: https://zeroerrors.github.io/flecs/

## Change Log
- Updated Doxygen to v1.10.0
  This adds many features and bug fixes.
  One of which is improved search. (See: https://github.com/doxygen/doxygen/pull/10218)
  | Before | After |
  |--------|-------|
  | image | image |
- Enabled Github style markdown links
  This changes the format of heading links from an incrementing number to a lower case version of the title. This will make links more stable when there are changes within the documentation. eg. adding new headings.
  **Before:** https://www.flecs.dev/flecs/md_docs_Manual.html#autotoc_md151
  **After:** https://zeroerrors.github.io/flecs/md_docs_2Manual.html#components-1
- Enabled Graphviz SVG graph generation
  This replaces the PNG generation done by Doxygen with SVG generation done by Graphviz resulting in higher quality graphs when rendered on high DPI displays.
- Updated Doxygen Awesome CSS
- Added opengraph and twitter metadata to the header html
- Fixed coredoc `URL_ROOT` creating invalid links due to `/` at the end of the URL. These links also now work thanks to using the Github markdown link style.

# Breaking Changes
There are two breaking changes this upgrade introduces.

### URLs to markdown documents has changed.
The path encoding changes from replacing `/` with `_` to replacing it with `_2`. (See: https://github.com/doxygen/doxygen/pull/9864) This changes the URLs for markdown files in the `docs/` directory from being `md_docs_<name>.html` to being `md_docs_2<name>.html`
**Before:** https://www.flecs.dev/flecs/md_docs_Docs.html
**After:** https://zeroerrors.github.io/flecs/md_docs_2Docs.html

### "Modules" has been renamed to "Topics" to avoid the name clashing with C++ Modules.
This only seems to change the one URL.
**Before:** https://www.flecs.dev/flecs/modules.html
**After:** https://zeroerrors.github.io/flecs/topics.html

# Issues
This is a pre-existing issue but there are links in some markdown files that use something such as `Queries.md#variables` this works on Github but does not work in Doxygen. There is an open issue about this problem. https://github.com/doxygen/doxygen/issues/9021 and https://github.com/doxygen/doxygen/issues/9701

### DIFF
--- a/.github/workflows/pages.yml
+++ b/.github/workflows/pages.yml
@@ -30,21 +30,26 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
-        
+
       - name: Setup Pages
         uses: actions/configure-pages@v4
-        
+
+      - name: Install Dependencies
+        run: |
+          sudo apt install --yes graphviz
+          wget https://github.com/doxygen/doxygen/releases/download/Release_1_10_0/doxygen-1.10.0.linux.bin.tar.gz
+          tar -xzf ./doxygen-1.10.0.linux.bin.tar.gz
+
       - name: Doxygen
-        uses: mattnotmitt/doxygen-action@v1.9.5
-        with:
-          doxyfile-path: 'docs/cfg/Doxyfile'
-        
+        run: |
+          ./doxygen-1.10.0/bin/doxygen docs/cfg/Doxyfile
+
       - name: Upload artifact
         uses: actions/upload-pages-artifact@v2
         with:
           # Upload entire repository
           path: 'docs/html'
-          
+
       - name: Deploy to GitHub Pages
         id: deployment
         uses: actions/deploy-pages@v3
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,4 @@ bazel-testlogs
 **/CMakeFiles/*
 .vs
 out
+docs/html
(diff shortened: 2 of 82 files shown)


SanderMertens:
Awesome! Just going through the docs at the moment, noticed that the (now) topic descriptions are gone:
Screenshot 2023-12-29 at 12 05 30 PM

which used to be:
Screenshot 2023-12-29 at 12 06 36 PM

Any idea what caused this?

ZeroErrors:
Not sure how I missed that. I did some digging still not sure why it is broken. The Doxygen codebase also isn't exactly easy to navigate so I haven't been able to debug it, but found that disabling `JAVADOC_AUTOBRIEF` seems to solve that issue but it means that it now no longer automatically generates the brief from the detailed description.

You can see the differences between:
| Before | After |
|--------|-------|
| image image | image image |

My guess is there is some bug in Doxygen that causes the `@brief` to not work with `@defgroup` when `JAVADOC_AUTOBRIEF` is enabled.

SanderMertens:
Hmm that's arguably worse... The missing descriptions for the defgroups are annoying though, the names by themselves are much less meaningful. Maybe there's a different way to provide the briefs for the defgroups now?

ZeroErrors:
Seems it works if we remove `@brief` and let the autobrief handle populating it.

There is one problem with the change which is that autobrief ends the brief when it finds the first `.` the full text is kept only in the detailed description.
This seems to be fine for all the current group documentation except for the Prefixes and Quantities which have `(e.g. ...)` so it gets cut off.
image

SanderMertens:
That's OK, the docs can be easily updated to not include dots. LGTM! Let me know if there's anything else you'd like to do, otherwise I'll merge it.

ZeroErrors:
I think its good to go. Can follow up afterwards if we find anything else that needs improvements.

PR #1109, opened 2024-01-01
Reduce copies when duplicating tables with non-POD components for snapshots

ZeroErrors:
I was browsing the code and noticed this case could be optimized.

1. Use `ecs_vec_init` then `ecs_vec_set_count` instead of `ecs_vec_copy`.
  Removes 1 copy.
2. Use `copy_ctor` instead of `ctor` then `copy`.
  Best case removes unnecessary initialization.
  Worse case no change.
  `ti->hooks.ctor` is always set if `ti->hooks.copy` is set, so we also remove a redundant branch.

Another thing I noticed is that the duplicated table could previously have empty space at the end since the vectors get size to the nearest power of 2, because snapshots don't get modified there isn't any need for this empty space so we could save some memory by only allocating vectors for the exact number of rows that is in the table.

This does mean that after restoring a snapshot adding an entity will cause table reallocation in order for it to grow. Perhaps it would be best if this was a flag that is passed when creating a snapshot so the user can chose the case they wish to optimize for?

### DIFF
--- a/flecs.c
+++ b/flecs.c
@@ -35124,7 +35124,7 @@ ecs_data_t* flecs_duplicate_data(
 
     ecs_data_t *result = ecs_os_calloc_t(ecs_data_t);
     int32_t i, column_count = table->column_count;
-    result->columns = flecs_wdup_n(world, ecs_column_t, column_count, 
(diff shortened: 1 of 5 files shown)


SanderMertens:
Thanks for the PR!

Issue #478, opened 2021-08-31
Incorrect memory alignment of components.


RobertBiehl:
**Describe the bug**
Component structs passed to system do not adhere to the defined memory alignment correctly.

**To Reproduce**

1. enable avx

2. Create component with aligned struct, e.g.
```
#include <cglm/cglm.h>
typedef struct EcsTransform {
    CGLM_ALIGN_MAT mat4 value;
} EcsTransform;
```

3. Inside system check address of pointer to struct

```
void TestSystem(ecs_iter_t* it) {
    EcsTransform3 *transform = ecs_term(it, EcsTransform3, 1);
    printf("(EcsTransform3 *) addr = %#010x", transform);

}
```
Sometimes prints addresses not aligned to 32 bytes as expected from CGLM_ALIGN_MAT.
This causes crashes in calls to AVX intrinsics.

**Expected behavior**
Address of component should be aligned correctly, in this case to 32 bytes.

**Additional context**
Add any other context about the problem here (operating system, hardware, ...).

SanderMertens:
Hm, flecs should be able to (and has been verified to) handle alignment correctly. It does rely on `alignof/ECS_ALIGNOF` reporting the correct alignment, could you verify that if you do `ECS_ALIGNOF(EcsTransform3)` it returns 32?

Additionally you could run the following code to check if the correct alignment has been registered with flecs:
```c
ECS_COMPONENT(world, EcsTransform3);

EcsComponent *ptr = ecs_get(world, ecs_id(EcsTransform3), EcsComponent);
printf("%d\n", ptr->alignment);
```

chengts95:
Yes, I have the same problem, I was thinking it was my fault, but the alignment of components is always 16bit even I changed the malloc function to use default 32bit alignment.

SanderMertens:
@RobertBiehl @chengts95 if you're interested in debugging this further, the part of the code that should take care of the alignment is `ecs_vector_t`. 

A vector is a contiguous block of memory with a header in front of it (containing the number of elements + allocated size). At an offset from the header the actual array begins, and it's important that this offset is aligned to the alignment of the component type.

The alignment is computed by macro's and passed into the vector functions (to allow for compile-time optimization where the vector type is known). If anything's wrong with how the component alignment is used, I would expect that to be in this code.

I'd like to look into this myself as well but it may take a bit of time before I can get to it.

chengts95:
[...]

PR #1110, opened 2024-01-02
Add alignment support to vec

ZeroErrors:
Fixes: https://github.com/SanderMertens/flecs/issues/478

# Changes
This change adds memory alignment support to vec which in turn fixes alignment support for components stored in tables.

Changes to vec.
1. Added `ecs_size_t alignment` argument alongside `ecs_size_t size` argument in vec API.
1. `ecs_vec_t` stores both `void* array` and `void* mem`.
  `array` is the same as before, it is the pointer to the data. `mem` is the original pointer that was allocated, this is used so we have the correct address to be able to realloc or free the memory.
1. Vec allocates a buffer of `size * elem_count + alignment` this is the `void* mem` pointer and then shifts the pointer in order to have the correct alignment for `void* array`.
1. When reallocating the vector a `memmove` may be necessary in order to correct for a different alignment of the new pointer.

This increases the memory usage of both the vec struct and the allocated arrays.

Since hashmap uses vec internally it also now has alignment support.

### DIFF
--- a/flecs.c
+++ b/flecs.c
@@ -637,7 +637,7 @@ typedef struct ecs_table__t {
     int32_t traversable_count;       /* Traversable relationship targets in table */
     uint16_t generation;             /* Used for table cleanup */
     int16_t record_count;            /* Table record count including wildcards */
-    
+
     struct ecs_table_record_t *records; /* Array with table records */
     ecs_hashmap_t *name_index;       /* Cached pointer to name index */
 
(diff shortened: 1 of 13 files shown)


SanderMertens:
Hm, I intentionally replaced the old vector code (which had alignment support) with a new simpler vector which simplified code & improved performance (likely due to the reduced memory footprint of table columns). I'm not sure if I want to add that back into the core.

My plan for supporting components with >16 byte alignment is pluggable storages, where you'll be able to implement your own data structures / allocation schemes.

ZeroErrors:
ah, yeah that makes sense.

PR #1111, opened 2024-01-03
Handle missing metatype

ZeroErrors:
Fixes a crash if a meta member is added that doesn't have `EcsMetaType`.

Previously in `flecs_meta_serialize_struct` when `flecs_meta_serialize_type` was called it would log a message and `return -1` directly after which it would call `op = flecs_meta_ops_get(ops, cur);` which would access an index out of bounds of the ops array.


SanderMertens:
Thanks for the PR!

PR #1115, opened 2024-01-05
Fill out ecs_iter_t.system field when iterating filters, queries, and rules

BeanCheeseBurrito:
Fills out the ``ecs_iter_t.system`` field with the entity id associated with the filter/query/rule if it has one.


SanderMertens:
Thanks for the PR!

PR #1122, opened 2024-01-08
docs: Link function names in doc comments.

waywardmonkeys:
This changes things to use `ecs_foo()` rather than `ecs_foo` so that doxygen's auto-linking will work.


waywardmonkeys:
Note that this only does function names (hopefully).

Other things will be done or fixed in subsequent passes to keep things easier to review (hopefully).

PR #1126, opened 2024-01-11
ecs_strbuf_t performance improvements, enable HTTP request caching from C API

SanderMertens:
This PR:
- Replaces the `ecs_strbuf_t` implementation with a simpler & more performant one
- Makes cache invalidation timeouts configurable for the HTTP server
- Enables caching for HTTP requests done from the C API (using `ecs_http_server_request`)


include/flecs/addons/http.h
```diff
@@ -126,6 +126,8 @@ typedef struct {
     uint16_t port;                    /**< HTTP port */
     const char *ipaddr;               /**< Interface to listen on (optional) */
     int32_t send_queue_wait_ms;       /**< Send queue wait time when empty */
+    ecs_ftime_t cache_timeout;             /**< Cache invalidation timeout (0 disables caching) */
+    ecs_ftime_t cache_purge_timeout;       /**< Cache purge timeout (for purging cache entries) */
```
waywardmonkeys, line 130: Is it worth noting that these are measured in seconds, and that the default for the `cache_purge_timeout` is 10x the `cache_timeout`?

test/meta/src/Serialized.c
```diff
@@ -2145,14 +2145,12 @@ void Serialized_ops_missing_metatype(void) {
...
     } X;
-    typedef struct {
-        X x;
-    } T;
 
     ecs_world_t *world = ecs_init();
 
     ECS_COMPONENT(world, X);
 
+    ecs_log_set_level(-4);
```
waywardmonkeys, line 2153: Is this intentional to still have here?

PR #1128, opened 2024-01-17
Use consistent doxygen syntax.

waywardmonkeys:
Many things used `@...` but some used `\...`, so use `@...` consistently.


SanderMertens:
Awesome, thanks for the PR!

PR #1133, opened 2024-01-29
docs: Fix more doxygen warnings.

waywardmonkeys:
Most of these are to correct unbalanced grouping. Two fixes are to correct a reference to a group name.


SanderMertens:
Thanks for the PR!

PR #1135, opened 2024-01-31
ci: Add some macOS-14 jobs on Apple Silicon.

waywardmonkeys:
`macos-latest` is currently `macos-12`. `macos-14` will become `macos-latest` in some months. `macos-14` runs exclusively on Apple Silicon hardware.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -193,7 +193,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


waywardmonkeys:
This doesn't (yet?) change the `build-macos` job which still runs on `macos-latest` (so `macos-12`).

A subsequent follow-up should be done to remove usage of `macos-11` builders as they will be removed later this year.

SanderMertens:
Thanks for the PR!

PR #1138, opened 2024-02-02
Docs snippet tabs

ZeroErrors:
## Change Log
- Added `flecs-snippet-tabs` a heavily modified version of `doxygen-awesome-tabs`.
  - Persists the last active tab
  - If a group of tabs doesn't have the current active tab it will use the last active tab it does have.
  This allows us to have a mix of tabs for `C`, `C++` and `Query DSL` (or others) and keep the active tab in sync without breaking other tab groups.
- Updated Systems docs to use tabs
  - Note in order for this to work correctly in Markdown for both Doxygen and GitHub it needs to be formatted in a specific way.
  
  Manually using HTML tags works around issues with how Doxygen generates HTML for markdown bullet lists. Adding spacing before the code block allow for GitHub to correctly display the code block in a bullet list.
  ``````md
  To manually run a system, do:
  <div class="flecs-snippet-tabs">
  <ul>
  <li><b class="tab-title">C</b>
  
  ```c
  ecs_run(world, ecs_id(Move), 0.0 /* delta_time */, NULL /* param */)
  ```
  </li>
  <li><b class="tab-title">C++</b>
  
  ```cpp
  flecs::system sys = ...;
  sys.run();
  ```
  </li>
  </ul>
  </div>
  ``````

- Fixed a couple minor issues in the systems docs.
  - Some C++ code marked as C
  - Fixed: `ecs_set_tasks_threads` -> `ecs_set_task_threads`
  - Two minor shifts in docs to make them group into the tabs better.
    - C++ System declaration
    - "Note how query iteration has an outer and an inner loop, ..."

Example: https://zeroerrors.github.io/flecs/md_docs_2Systems.html


SanderMertens:
This is really cool 🥇 Thanks for the PR!

PR #1139, opened 2024-02-02
c++: Mark all explicit move ctors and operator= as `noexcept`.

waywardmonkeys:
Defaulted move ctors and operator= will be `noexcept` when possible automatically.


SanderMertens:
Thanks for the PR!

PR #1147, opened 2024-02-13
Improve robustness when deserializing JSON into  worlds with mismatching reflection data

SanderMertens:
This PR implements the following features & bugfixes:
- Order of component registration no longer needs to match between serialization/deserialization
- Fix issues with deserializing entities/members with large names (>256 characters)
- Fix off-by-one issue with deserializing strings that are exactly 256 characters
- Ignore serialized values for components for which no/mismatching reflection data exists
- Introduce new strict mode that enables the old behavior
- Implement check that ensures a type with mismatching reflection data only throws one error (vs. one per value)
- Add `ecs_world_from_json_file` / `world::from_json_file` function that loads JSON data from file
- Significantly reduce creation of empty tables during deserialization

### DIFF
--- a/include/flecs/addons/json.h
+++ b/include/flecs/addons/json.h
@@ -41,6 +41,10 @@ typedef struct ecs_from_json_desc_t {
         const char *value,
         void *ctx);
     void *lookup_ctx;
+
+    /** Require components to be registered with reflection data. When not
+     * in strict mode, values for components without reflection are ignored. */
+    bool strict;
 } ecs_from_json_desc_t;
 
 /** Parse JSON string into value.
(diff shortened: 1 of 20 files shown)


include/flecs/addons/json.h
```diff
@@ -84,13 +88,28 @@ const char* ecs_entity_from_json(
...
+ * @param desc Deserialization parameters.
+ * @return Last deserialized character, NULL if failed.
  */
 FLECS_API
 const char* ecs_world_from_json(
     ecs_world_t *world,
     const char *json,
     const ecs_from_json_desc_t *desc);
 
+/** Same as ecs_world_from_json, but loads JSON from file.
```
waywardmonkeys, line 100: This should be `ecs_world_from_json()` with the parentheses so that doxygen links it correctly.

waywardmonkeys (changes requested):

include/flecs/addons/json.h
SanderMertens: Fixed!

PR #1151, opened 2024-02-16
Changed enum constants to keep contiguous array and modifies index/search strategy

garrett-is-a-swann:
This modifies the storage of enum reflection data from a sparse array lookup table to a contiguous-array with both lookup table/linked list scanning strategies.

Changes enum constants array constant (default) size of 128 to a compile-time determined size based on number of enum constants found during reflection.

Modified enum reflection search mechanisms by:
- Finds 0->128 in ascending order.
- Generates 0->128 by recursively dividing range, reducing the template depth from n to ln(n).
- Additionally finds bitmasks from `1 << (sizeof(underlying_type) * 8 - 1)` until it reaches the value used by the previous search (128).

It maintains a `contiguous_until` index, which is used to track the last default enum value. This is used as a heuristic to shortcut iteration, and direct lookup in the case that the user's enum is using default enumeration values for part or all of the enumerators. If the enum is sparse or the value is beyond the contiguous section, `contiguous_until` index is skipped to and a search on reflection data occurs instead.

### DIFF
--- a/flecs.h
+++ b/flecs.h
@@ -16726,34 +16726,56 @@ struct enum_reflection {
         return static_cast<underlying_type_t<E>>(Value != from_int<0>());
     }
 
-    template <E Low, E High, typename... Args>
-    static constexpr underlying_type_t<E> each_enum_range(underlying_type_t<E> last_value, Args... args) {
+    template <E Low, E High, typename Args>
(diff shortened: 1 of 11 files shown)


SanderMertens:
Awesome, LGTM!

PR #1154, opened 2024-02-22
Remove `FLECS_LEGACY`.

waywardmonkeys:
This was to enable compiling with c89, but this is no longer functioning as the code is using c99.

### DIFF
--- a/flecs.h
+++ b/flecs.h
@@ -45,11 +45,6 @@
 #define ecs_ftime_t ecs_float_t
 #endif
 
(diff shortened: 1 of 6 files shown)


SanderMertens:
I want to keep this macro- though will have to add tests to make sure it works correctly, I don't think it currently does.

The intent of the macro is not to compile flecs as C89, but to exclude any definitions in the public headers that require C99 (like variadic macro's). That way C89 projects can link with a flecs library that's compiled with a C99 compiler.

PR #1155, opened 2024-02-22
Use `fseek(file, 0, SEEK_SET)` instead of `rewind(file)`.

waywardmonkeys:
This removes a warning when using `clang-tidy` as `rewind()` can't report errors while `fseek()` can. (Not that we're checking for errors here anyway...)


SanderMertens:
LGTM!

PR #1156, opened 2024-02-22
ci: Remove macOS-11 builds, add macOS-13.

waywardmonkeys:
The macOS 11 build runners are being deprecated and are intended to be removed by the middle of 2024:

https://github.com/actions/runner-images/issues/9255

Additionally, macOS-12 is to be deprecated in Q3 and removed in Q4, so prepare for that now by adding a macOS-13 build.


SanderMertens:
LGTM

PR #1158, opened 2024-02-24
Add overloads for flecs::entity and depends_on to allow scoped enums with an underlying type of entity_t.  

RoyAwesome:
I have some enums like:

```cpp
 enum class engine : flecs::entity_t
 {
     main_window = EcsFirstUserEntityId,
     input,
     renderer,
     scripting,
     MAX,
 };
```

which allows me to just add entity ids in sequence for hardcoded entities.  However, none of the flecs functions accept passing these enums as entity_t.  I have to cast to the underlying type in my code. 

So, I added some functions to accept enums like this so that I can just use these enums as I want.  

```cpp
//he::entities::pipeline is an enum with underlying type of flecs::entity_t
 world.entity(he::entities::pipeline::on_render_start).add(flecs::Phase).depends_on(he::entities::pipeline::on_window_start);
```

I only grabbed a few functions I'm currently using.  If this PR is accepted, I will probably add more as I go through.

### DIFF
--- a/include/flecs/addons/cpp/entity.hpp
+++ b/include/flecs/addons/cpp/entity.hpp
@@ -84,9 +84,10 @@ struct entity : entity_builder<entity>
      *
      * @param id The enum with an underlying type of flecs::entity_t value to convert.
      */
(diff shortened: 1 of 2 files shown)


RoyAwesome:
closed because #1151 is way better than this

PR #1160, opened 2024-02-26
Additional overloads for enum entities

RoyAwesome:
With the addition of using enums as entity ids, there are a few places where the API takes an flecs::entity_t, but not an enum.  I've added a few of those cases.


SanderMertens:
LGTM!

PR #1165, opened 2024-03-05
Add EcsModule to component-parents

garrett-is-a-swann:
Add EcsModule to component parents if they exist. This is to avoid deletion of namespace entities during world fini, which can lead to component deletion before observers have a chance to fire.


SanderMertens:
LGTM!

PR #1168, opened 2024-03-11
Added cpp entity.emplace<T>(...) for T(flecs::entity, Args...) constructors

garrett-is-a-swann:
It looks like this mechanism of entity.emplace<T>(...):
```
     * Emplace attempts the following signatures to construct the component:
     *
     * @code
     * T{Args...}
     * T{flecs::entity, Args...}
     * @endcode
     *
     * If the second signature matches, emplace will pass in the current entity 
     * as argument to the constructor, which is useful if the component needs
     * to be aware of the entity to which it has been added.
```
was removed here: https://github.com/SanderMertens/flecs/commit/75c43f22dc4096bf170776225f7924acaae117d0?diff=split&w=0

The goal of this PR is to restore this functionality.


SanderMertens:
Closing this (discussed on Discord)

PR #1170, opened 2024-03-12
Added cpp entity.emplace<T>(...) for static T::build(flecs::entity, Args...) builders

garrett-is-a-swann:
This mechanism for entity.emplace<T>(...) was removed noted in the comments:
```
     * Emplace attempts the following signatures to construct the component:
     *
     * @code
     * T{Args...}
     * T{flecs::entity, Args...}
     * @endcode
     *
     * If the second signature matches, emplace will pass in the current entity 
     * as argument to the constructor, which is useful if the component needs
     * to be aware of the entity to which it has been added.
```
here: https://github.com/SanderMertens/flecs/commit/75c43f22dc4096bf170776225f7924acaae117d0?diff=split&w=0

The goal of this PR, like https://github.com/SanderMertens/flecs/pull/1168, is to restore this functionality but by locating a free-floating builder function on the type for the building. This avoids `this` invalidation.


SanderMertens:
Closing this (discussed on Discord)

PR #1176, opened 2024-03-21
fix memory read/write access in out of bounds memory on copying comps

Indra-db:
fix memory read/write access in out of bounds memory on copying components when using ECS operations where count > 1, such as nested_prefabs

without this fix, the newly introduced test case would segfault. This bug happens due to the fact that in C you loop count times, but also offset the dest_ptr, and then within the copy impl of C++, it loops count again, this means you would go count-1 * size_obj out of memory bounds for src as well as dest ptr.

This fix is the correct fix as it limits src ptr to just 1, while the dest ptr still gets offset each iteration.

(this was previously discussed with sanders, the information above is just for tracking why & what)

### DIFF
--- a/flecs.c
+++ b/flecs.c
@@ -15148,7 +15148,7 @@ void flecs_override_copy(
     int32_t i;
     if (copy) {
         for (i = 0; i < count; i ++) {
-            copy(ptr, src, count, ti);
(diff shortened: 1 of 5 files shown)


SanderMertens:
LGTM!

PR #1182, opened 2024-04-08
Fix warnings when `FLECS_USE_OS_ALLOC`.

waywardmonkeys:
When using the OS allocator, `flecs_balloc_block()` is unused and can be removed. Also, the `src` parameter to `flecs_brealloc()` is not used.


SanderMertens:
LGTM!

PR #1183, opened 2024-04-10
Fix an assert when moving a component without a ctor to an existing table

Thinkofname:
This changes the code in `flecs_table_merge_column` to not have `flecs_table_grow_column` default construct components in the new rows added to the column and instead handle it via `ctor_move_dtor` instead. This allows components that cannot be default constructed to be moved into existing tables.


SanderMertens:
LGTM!

PR #1184, opened 2024-04-13
use destructive move semantics when move & move_ctor is not set. 

Indra-db:
If neither move nor move_ctor are set, this indicates that non-destructive move 
semantics are not supported for this type. In such cases, we set the move_dtor
as ctor_move_dtor, which indicates a destructive move operation. 
This adjustment ensures compatibility with different language bindings.

### DIFF
--- a/src/storage/table.c
+++ b/src/storage/table.c
@@ -1668,10 +1668,10 @@ void flecs_table_delete(
 
                 ecs_move_t move_dtor = ti->hooks.move_dtor;
                 
-                // If move_ctor is not set but ctor_move_dtor is set, assign move_dtor to ctor_move_dtor.
-                // This adjustment is crucial for compatibility across different language bindings where
-                // the standard memory model of C may not apply, potentially altering how operations
-                // within the table are handled.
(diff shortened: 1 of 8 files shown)


src/storage/table.c
```diff
@@ -1667,6 +1667,16 @@ void flecs_table_delete(
                 }
 
                 ecs_move_t move_dtor = ti->hooks.move_dtor;
+                
+                // if move_ctor & move is not set and ctor_move_dtor & move_dtor is set,
+                // use move_dtor as ctor_move_dtor. The reason we
```
copygirl, line 1672: Shouldn't this say "use ctor_move_dtor as move_dtor"?

Indra-db: should it? since we are using the "move_dtor", but it's using underneath "ctor_move_dtor" , hence as? 

I might be wrong, it's definitely a word twist

Indra-db: committed a change to the comment due to the word twist.

SanderMertens:
LGTM! (CI fail is due to an issue that's already fixed on master)

PR #1185, opened 2024-04-13
change entity constructor to take entity_t instead of id_t

Indra-db:
entity's should not represent possibly pairs which are a possibility with id_t. 
Hence the change.


SanderMertens:
LGTM!

PR #1188, opened 2024-04-17
[Cpp] add missing get mut functions to world

Indra-db:


SanderMertens:
LGTM! Can you add a test case for it?

Indra-db:
@SanderMertens done

PR #1189, opened 2024-04-18
Remove inexistent ecs_meta_set_component

pfeodrippe:
I am having link issues and couldn't find where this function is defined in this project, let me know if I'm not seeing something \o

### DIFF
--- a/flecs.c
+++ b/flecs.c
@@ -45341,10 +45341,10 @@ void flecs_table_delete(
 
                 ecs_move_t move_dtor = ti->hooks.move_dtor;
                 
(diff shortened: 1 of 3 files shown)


SanderMertens:
Nice catch! Can you run `bake` once on the flecs repository so that `flecs.h` also gets updated?

pfeodrippe:
> Nice catch! Can you run `bake` once on the flecs repository so that `flecs.h` also gets updated?

Done, thanks!

SanderMertens:
LGTM!

PR #1190, opened 2024-04-19
Add pair overload to get_ref

TBlauwe:
First time creating a pull request, I'm sorry in advance if I did something the wrong way !

This pull request adds a pair overload to `get_ref` to support pair types, so that a ref to the actual type is returned rather than a ref to the pair.

Here is a repro : 

```cpp
struct A {};
struct B {int value {0};};
using AB = flecs::pair<A, B>;
using SharedAB = flecs::pair<A, flecs::ref<B>>;

int main(int argc, char **argv)
{
    flecs::world world;
    auto a = world.entity("a");
    auto b = world.entity("b");
    a.add<AB>();
    b.set<SharedAB>({a.get_ref<AB>()});// This doesn't work. 
                                       // Returns flecs::ref<flecs::pair<A, flecs::ref<B>>>
    b.set<SharedAB>({a.get_ref<A, B>()}); // A workaround
                                          // Returns flecs::ref<B>
}
```

Changes were made to the normal `get_ref` to dissambiguate with the newly added overload.

### DIFF
--- a/include/flecs/addons/cpp/entity.hpp
+++ b/include/flecs/addons/cpp/entity.hpp
@@ -225,13 +225,30 @@ struct entity : entity_builder<entity>
      * @tparam T component for which to get a reference.
      * @return The reference.
      */
-    template <typename T>
+    template <typename T, if_t< is_actual<T>::value > = 0>
     ref<T> get_ref() const {
         return ref<T>(m_world, m_id, _::cpp_type<T>::id(m_world));
     }
 
+    /** Get reference to component.
+     * Overload for when T is not the same as the actual type, which happens
+     * when using pair types.
+     * A reference allows for quick and safe access to a component value, and is
+     * a faster alternative to repeatedly calling 'get' for the same component.
+     *
+     * @tparam T component for which to get a reference.
+     * @return The reference.
+     */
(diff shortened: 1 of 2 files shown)


SanderMertens:
Looks good! Can you add a test that makes sure the right pair gets added?

The CI failure is not due to this PR, investigating at the moment.

TBlauwe:
Done ! Although I may have misunderstood you. Should I also add these ?

```cpp
// In Entity.cpp
auto e = world.entity().set<PositionTag>({10, 20});
test_bool((e.has<Position, Tag>()), true);
test_bool((e.has<Tag, Position>()), false);
// ....
auto e = world.entity().set<TagPosition>({10, 20});
test_bool((e.has<Position, Tag>()), false);
test_bool((e.has<Tag, Position>()), true);
 ```
 
 If so, can I add these two `using` in cpp_api.h ?
 
 ```cpp
 // In cpp_api.h
using PositionTag = flecs::pair<Position, Tag>;
using TagPosition = flecs::pair<Tag, Position>;
 ```

copygirl:
Would it make more sense for the test and reader of the test to mix setting `<PositionTag>` with getting via `<Position, Tag>` (for example), both to make sure they mean the same thing to the library, and for the reader to see these are equivalent?

PR #1193, opened 2024-04-20
Fix duplicate definition errors when using unity builds

apache-hb:
This guards a couple of internal headers to prevent duplicate definition errors when unity builds are used via meson


SanderMertens:
LGTM for now- is there a way this can be reproduced in CI? v4 is going to add a lot of new internal headers that might break your build (though I'm not 100% sure how it's failing).

apache-hb:
> is there a way this can be reproduced in CI?

Building with `meson setup build -Dunity=true -Dunity_size=200` will probably catch some errors as they pop up, unfortunately unity builds are a bit hard to reproduce perfectly.
> v4 is going to add a lot of new internal headers that might break your build.

I already follow the master branch and keep it up to date, I'd be happy to make another pr once v4 releases to fix any issues that might have cropped up.
> (though I'm not 100% sure how it's failing).

Unity builds cause multiple translation units to get merged together, in the case of unguarded internal headers this means symbols that were meant to be private to a single TU can leak into another TU that also defines the same symbols, causing the multiple definition errors.

SanderMertens:
Ah ok, I'll see if that can be added to CI. Thx for the background!

PR #1194, opened 2024-04-20
Prevent buffer overflow with large compiler version strings

### LINKED ISSUE #1192: flecs crashes at startup when using clang built from git
**Describe the bug**
when using clang built from source flecs crashes because the `__clang_version__` string overflows `flecs_compiler_info`

apache-hb:
Statically computes the compiler version string rather than determining it at runtime. resolves #1192


SanderMertens:
LGTM!

PR #1202, opened 2024-05-01
CXX-20 MODULES - global module fragment support

alexv-ds:
This PR is based on one sleepless night and a similar [PR for entt](https://github.com/skypjack/entt/pull/1047).

The problem is that we get compilation errors when using the methods (when include flecs.h in global module fragment):
```cpp
world.import<ModuleName>();
world.system<ComponentName>();
// And perhaps others that call some template functions, the corrections of which are given in this PR
```

The reason to be that certain function declarations that are implicitly required by the implementation aren't exported into the module interface since they are marked 'static' and have internal linkage.

And when compiling we get the following error (clang-18 ubuntu 24.04 package version 1:18.1.3-1):
```console
# This is an incomplete output for readability purposes.
/home/alexv/Desktop/flecs/include/flecs/private/../addons/cpp/mixins/term/../../utils/signature.hpp:44:25: error: no matching function for call to 'type_to_inout'
   44 |             , inout ({ (type_to_inout<Components>())... })
      |                         ^~~~~~~~~~~~~~~~~~~~~~~~~
```

This pull request solves this problem. I also advise to look at the original pull request for entt (link at the beginning of the description)

### DIFF
--- a/include/flecs/addons/cpp/component.hpp
+++ b/include/flecs/addons/cpp/component.hpp
@@ -30,7 +30,7 @@ namespace _ {
 
 #if defined(__GNUC__) || defined(_WIN32)
 template <typename T>
-inline static const char* type_name() {
+inline const char* type_name() {
     static const size_t len = ECS_FUNC_TYPE_LEN(const char*, type_name, ECS_FUNC_NAME);
     static char result[len + 1] = {};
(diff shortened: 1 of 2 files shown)


SanderMertens:
LGTM!

PR #1210, opened 2024-05-27
Added Enum specialization for event_builder_base::id(Enum::Value)

garrett-is-a-swann:
Adds `event_builder_base::id(Enum::Value)` specialization and companion test.


garrett-is-a-swann:
Opened #1224 for PR into v4 branch.

PR #1211, opened 2024-05-28
Fix issue where the use of a function pointer cast can trip -fsanitize=undefined on recent versions of clang

darkuranium:
**Note:** The changes were made against v3.11.2, *not* master. Hopefully they'll be useful anyway.

---

The error in question: (plus others, for each case of a casted dtor function pointer)

```
<...>/flecs/src/bootstrap.c:142:8: runtime error: call to function flecs_observer_fini through pointer to incorrect function type 'void (*)(void *)'
<...>/flecs/src/observer.c:1017: note: flecs_observer_fini defined here
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior <...>/flecs/src/bootstrap.c:142:8 
```

### DIFF
--- a/src/addons/rules/api.c
+++ b/src/addons/rules/api.c
@@ -103,6 +103,13 @@ void flecs_rule_fini(
     ecs_poly_free(rule, ecs_rule_t);
 }
 
+/* ecs_poly_dtor_t-compatible wrapper */
+static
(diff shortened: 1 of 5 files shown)


SanderMertens:
Thanks for the PR!

PR #1213, opened 2024-06-12
Ensure Windows version macro is minimum required to include inet_pton

benjitrosch:
This should fix a _very obscure_ issue I was running into when cross-compiling from WSL using `x86_64-w64-mingw32-gcc-posix`, where WINVER was not defined and _WIN32_WINNT was 0x502 (Windows Server 2003) which was not a recent enough version to use `inet_pton` which caused some issues with #include's.

Solution based on this article from Microsoft which details how to manually set these values: https://learn.microsoft.com/en-us/cpp/porting/modifying-winver-and-win32-winnt?view=msvc-170


SanderMertens:
Thanks for the PR!

PR #1219, opened 2024-06-19
Add term_index method to cpp iterator

jpeletier:
As per [our discussion in Discord](https://discord.com/channels/633826290415435777/1032491261548044299/threads/1237529263733735524), here is a convenience method to access what term triggered an observer, along with a test.


SanderMertens:
Thanks for the PR!

PR #1224, opened 2024-07-03
[v4] Added Enum specialization for event_builder_base::id(Enum::Value)

garrett-is-a-swann:
Adds ability to create events with enum component ids via `event_builder_base::id(Enum::Value)` specialization, and adds companion test.

Replaces #1210 (master PR) for v4 branch


SanderMertens:
Thanks for the PR!

PR #1228, opened 2024-07-06
Update Quickstart observer to use 'events' instead of 'event'

lenis0012:
The latest release of flecs uses an array for events in the description of observer


SanderMertens:
Thanks for the PR!

PR #1236, opened 2024-07-10
Add better cs syntax support

Indra-db:
before:
image
after:
image


waywardmonkeys:
Is the C# support in highlightjs not so good? (Is that what's in the first screenshot or is that what Oxygen shows?)

Indra-db:
> Is the C# support in highlightjs not so good? (Is that what's in the first screenshot or is that what Oxygen shows?)

yes, not good. (We discussed it just now on work call)

PR #1262, opened 2024-07-11
Add support for generic callables to delegates

### LINKED ISSUE #1233: Support generic lambdas as `each` callbacks
## Problem

Implementing systems with an `each` callback requires repeating the component types. For example:

```cpp
world.system<FirstDescriptiveComponentName, const SecondDescriptiveComponentName>()
  .each([](FirstDescriptiveComponentName &a, const SecondDescriptiveComponentName &b) {
    // ...
  });
```

## Solution

It would be cleaner to use `auto` and pass a generic lambda to `each`.

```cpp
world.system<FirstDescriptiveComponentName, const SecondDescriptiveComponentName>()
  .each([](auto &a, auto &b) {
    // ...
  });
```

This doesn't work because the type of the `operator()` function is used to determine the arity (via `function_traits`). Directly determining the arity of a generic function isn't really possible (because what's the arity of `auto...`?). Instead, `each_delegate` could check if the given function can be called with a set of arguments. For the entity + components version, that could look like this:

```cpp
    template <template<typename X, typename = int> class ColumnType, 
        typename... Args, if_t< 
            sizeof...(Components) == sizeof...(Args)> = 0>
    static auto invoke_callback(
        ecs_iter_t *iter, const Func& func, size_t, Terms&, Args... comps)
    -> decltype(func(
        std::declval<flecs::entity>(),
        (std::declval<ColumnType<remove_reference_t<Components>>>().get_row())...
    ), void())
```

I made this change to my local copy and it seemed to work OK for me. This is inspired by EnTT's `each` which supports generic lambdas.

## Wider change

There are other places where this sort of change could be beneficial. All places that check for function arity in the same way are candidates. `iter` and `find` are another two examples.

indianakernick:
Updates several places that call a user-provided callable to select overloads using SFINAE rather than determining function arity using the callable's type. This allows generic lambdas to be used in places where they previously weren't allowed.

Resolves #1233.

### DIFF
--- a/flecs.h
+++ b/flecs.h
@@ -25735,7 +25735,7 @@ private:
         if_t<sizeof...(Components) == sizeof...(Args)> = 0,
         decltype(bool(std::declval<const Fn&>()(
             std::declval<flecs::entity>(),
-            std::declval<ColumnType< remove_reference_t<Components> > >().get_row()...)), 0) = 0>
+            std::declval<ColumnType< remove_reference_t<Components> > >().get_row()...))) = true>
     static flecs::entity invoke_callback(
         ecs_iter_t *iter, const Func& func, size_t, Terms&, Args... comps) 
     {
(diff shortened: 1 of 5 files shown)


SanderMertens:
Can you target the v4 branch with this PR? I'm days away from merging it, will reduce merge conflicts down the line.

SanderMertens:
Awesome, thanks for the PR!

PR #1265, opened 2024-07-12
Improve entity builder docs.

waywardmonkeys:
This fixes cross-linking for existing `@see` stuff by supplying the `const` modifier.

An instance of trying to link to `override` instead of `auto_override` was adjusted.

It adds some missing parameter documentation.

A line wrapping mishap in `emplace_auto_override` was also corrected.


SanderMertens:
Thanks for the PR!

PR #1266, opened 2024-07-12
docs(systems): Rust: Fix referring to wrong component.

waywardmonkeys:
The code should refer to `Transform` as the other languages and docs do.


SanderMertens:
Thanks for the PR!

PR #1273, opened 2024-07-15
Doxyfile: Undo search/replace, use 'Filters' term

waywardmonkeys:


SanderMertens:
I actually think this should be either rules or filters :-) If you see querys, it's a filter/rule find&replace renaming artifact.

waywardmonkeys:
@SanderMertens Right! Fixed and re-pushed.

The good news is that with this, there are no known spelling errors in the codebase. :)

SanderMertens:
Thanks for the PR!

PR #1278, opened 2024-07-19
[v4] Fix crash when copying world cpp object during fini

jpeletier:
When inadvertently copying a `flecs::world` object during shutdown, e.g., as part of a component `on_remove` hook, `ecs_fini()` is re-entered causing an abort.

Reproduction: 

```cpp
  struct A {
    int a;
  };

  flecs::world ecs;

  // declare on remove hook for component A:
  ecs.component<A>().on_remove([](flecs::entity e, A &) {  //
    // This code runs on world destroy, since we did not remove this component manually
    // before the world was destroyed.
    
    // obtain the entity's world. This increments the hdr refcount when it already was zero
    flecs::world ecs_copy = e.world();

    // here ecs_copy object wrapping c world is destroyed, and `world::~world()` for this copied instance
    // will be called, causing a repeated call to `ecs_fini()`
    // therefore, world destroy will be called again wreaking havoc.
  });

  ecs.entity().add<A>();

  // world will be destroyed here. Program will crash when `ecs_copy` goes out of scope as part
  // of the on_remove hook handler.

```

This PR corrects this issue and adds a test to verify the above code works as expected.

### DIFF
--- a/flecs.h
+++ b/flecs.h
@@ -20504,6 +20504,10 @@ struct world {
                 if (ecs_stage_get_id(world_) == -1) {
                     ecs_stage_free(world_);
                 } else {
+                    // before we call ecs_fini(), we increment the reference count back to 1
+                    // otherwise, copies of this object created during ecs_fini (e.g. a component on_remove hook)
+                    // would call again this destructor and ecs_fini().
+                    flecs_poly_claim(world_);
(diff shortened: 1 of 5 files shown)


SanderMertens:
Thanks for the PR!

PR #1283, opened 2024-07-29
add a constructor for flecs::ref<T> that just takes the entity

RoyAwesome:
flecs::ref<T> is a really useful object, but it's constructor is a bit weird.  It only takes a world_t* and an entity_t, both are encapsulated in the flecs::entity struct.  

So I added a constructor that simply takes a flecs::entity and extracts the world_t* and entity_t from that.  

There was a slight issue adding it because of how the flecs.hpp includes were laid out.  flecs::world has a get_ref<T>() which returns singletons, and expects ref to be declared before it.  I simply forward declared it in flecs.hpp like component is, and it seemed to like that.  

Also I'm kinda yoloing this PR cause I dont have bake setup on this new dev environment.  Hopefully the tests pass :)

Ok, tests passed, and I slipped in an operator bool() overload so 

```cpp
flecs::ref<Position> r(e);
if(r)
{
  //...
}
```

works.

### DIFF
--- a/flecs.h
+++ b/flecs.h
@@ -16457,6 +16457,9 @@ struct untyped_component;
 template <typename T>
 struct component;
 
+template <typename T>
+struct ref;
+
 namespace _
 {
 template <typename T, typename U = int>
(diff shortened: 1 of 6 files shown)


SanderMertens:
Thanks for the PR!

PR #1284, opened 2024-07-29
fix log tag

kagetu:
fix log tag


SanderMertens:
The reason why the tests for this PR fail is because flecs.c isn't updated: https://github.com/SanderMertens/flecs/blob/master/flecs.c

kagetu:
> The reason why the tests for this PR fail is because flecs.c isn't updated: https://github.com/SanderMertens/flecs/blob/master/flecs.c

finally, all checks have passed.
Sorry for having submitted the same PR many times. I just noticed that I can modify it in the previous PR. I mistakenly understood 'close PR' as meaning 'delete' or 'cancel'. :(
apologize once more for this PR.

PR #1287, opened 2024-08-01
yield_existing: Examples refer to `EcsIterable`

waywardmonkeys:
This feature was removed in v4, so shouldn't be mentioned in the examples.


SanderMertens:
Thanks for the PR!

PR #1293, opened 2024-08-06
add option to explicitly use provided offset, even when 0.

Indra-db:
This allows to register fields in arbitrary order for reflection. This is useful for example where in Rust fields are not guaranteed to be in the same order as represented.

### DIFF
--- a/include/flecs/addons/meta.h
+++ b/include/flecs/addons/meta.h
@@ -202,7 +202,7 @@ typedef struct EcsMember {
     int32_t count;                                 /**< Number of elements (for inline arrays). */
     ecs_entity_t unit;                             /**< Member unit. */
     int32_t offset;                                /**< Member offset. */
-    bool explicit_offset;                          /**< if Offset should be explicitly used. */
+    bool explicit_offset;                          /**< If offset should be explicitly used. */
 } EcsMember;
 
 /** Type expressing a range for a member value */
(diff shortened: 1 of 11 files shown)


include/flecs/addons/meta.h
```diff
@@ -202,6 +202,7 @@ typedef struct EcsMember {
     int32_t count;                                 /**< Number of elements (for inline arrays). */
     ecs_entity_t unit;                             /**< Member unit. */
     int32_t offset;                                /**< Member offset. */
+    bool explicit_offset;                          /**< if Offset should be explicitly used. */
```
copygirl, line 205: Should probably also get the capitalization fix.

Indra-db: yup, noticed when the CI failed haha :D, thanks !

Indra-db:
> Looks good! Can you:
> 
> * Add a test that shows the new feature working (out of order member declaration with 0 offset not being the first member)
> * Rename the member to `use_offset`

added test, renamed and fixed `.member` setting the field correctly with overloading.

include/flecs/addons/meta.h
```diff
@@ -202,6 +202,7 @@ typedef struct EcsMember {
     int32_t count;                                 /**< Number of elements (for inline arrays). */
     ecs_entity_t unit;                             /**< Member unit. */
     int32_t offset;                                /**< Member offset. */
+    bool use_offset;                          /**< If offset should be explicitly used. */
```
copygirl, line 205: Another nitpick: Alignment is off.

SanderMertens:
Thanks for the PR!

PR #1294, opened 2024-08-06
chore: create .gitattributes

epreston:
preserve line endings when building and testing on windows.


epreston:
This stops git from listing 50 files as modified due to line endings.
Reproduce by running `bake rebuild` and `bake test` on windows.

SanderMertens:
Thanks for the PR!

PR #1296, opened 2024-08-08
Migrate flecs script tutorial to v4

awest03:
I've started migrating the flecs script tutorial over to flecs v4 syntax and using the new explorer, the screenshots were taken with GNOME Web in dark mode which I think looks clean (a bit cleaner than my Firefox setup).

Some things which I've changed (I'll update this as I go):
- The images no longer hyperlink to the explorer in the state of the image (I gather the v4 explorer doesn't support this currently)


SanderMertens:
> The images no longer hyperlink to the explorer in the state of the image (I gather the v4 explorer doesn't support this currently)

Yep, that's what was blocking me to port the review over. I need to reimplement script loading from the URL.

awest03:
Okay, I'll try to continue changing it over to v4 scripts and we can put the links back in once the explorer supports it. Is it likely we would need to change the screenshots again at that point?

SanderMertens:
The screenshots can stay the same :) will just be a link update.

awest03:
I haven't updated this lately, I have completed up to the templates section, however there is a bug in flecs-game where when generating the grid from a Grid component the entities aren't added as children. I believe it's due to the changes in v4 related to `ecs_set_scope`, `ecs_new` and `ecs_new_w_pair` which are used to generate the grid

PR #1299, opened 2024-08-11
Fix C++ API observers with sparse components by filling out iterator field flags in flecs_uni_observer_invoke

Megafunk:
I noticed C++ observers with sparse components cause this assert to fire: 

`fatal: flecs.c: 10947: assert: !(idr->flags & EcsIdIsSparse) use ecs_field_at to access fields for sparse components (INVALID_OPERATION)`

A simple repro:

```cc
#include "flecs.h"
#include <iostream>

int main(int, char *[])
{
    struct Position {
        double x, y;
    };
    
    flecs::world ecs;
    ecs.component<Position>().add(flecs::Sparse);

    ecs.observer<Position>().event(flecs::OnSet).each([](flecs::entity e, Position& Pos)
    {
       std::cout << " - Position: {" << Pos.x << ", " << Pos.y << "}\n";
    });

    for (int i = 0; i < 4; ++i)
    {
        flecs::entity MeshEntity1 = ecs.entity();
        MeshEntity1.set<Position>({1,2});
    }
    
    ecs.progress();
}
```

After some messing around I found that `each_delegate::invoke(ecs_iter_t *iter)` seems to expect `ref_fields` to be set to determine if it uses `terms.populate` rather than `terms.populate_self`. I am not sure if this is the correct fix as this is the first time I've really done anything in flecs internals.

I think it would be useful to run the test cases with different combinations of sparse and non-sparse components to catch things like this ahead of time for commonly used API things.

### DIFF
--- a/flecs.c
+++ b/flecs.c
@@ -14093,8 +14093,10 @@ void flecs_uni_observer_invoke(
     it->callback_ctx = o->callback_ctx;
     it->run_ctx = o->run_ctx;
     it->term_index = impl->term_index;
-    it->query = o->query;
-
+    it->query = query;
+    it->ref_fields = query->fixed_fields | query->row_fields;
+    it->row_fields = query->row_fields;
+    
     ecs_entity_t event = it->event;
     int32_t event_cur = it->event_cur;
     it->event = flecs_get_observer_event(term, event);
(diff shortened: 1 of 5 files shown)


SanderMertens:
Nice! Can you add a test case for the issue as well?

Megafunk:
I added a single test for sparse on set observers based on the others in the file

SanderMertens:
Thanks for the PR!

PR #1304, opened 2024-08-15
correctly call populate self on query term building (CPP)

Indra-db:


Indra-db:
removed 
```fields_[index].is_ref = iter->sources[index] != 0;
        populate_self(iter, index + 1, comps ```
      originally thought it would need it, but since populate_self only can give false, it's useless to set.

SanderMertens:
Thanks for the PR!

PR #1305, opened 2024-08-15
simplify query sparse population + improve performance

Indra-db:
I'm not 100% sure on the changes since from my experimentation row == index in all cases, but I'm not sure if that's actually always the case? Tests do pass, but I'm not sure if there's enough tests covering this.

### DIFF
--- a/flecs.h
+++ b/flecs.h
@@ -25619,11 +25619,9 @@ private:
             if_not_t< is_empty<A>::value > = 0>
     void populate(const ecs_iter_t *iter, size_t index, T, Targs... comps) {
         if (iter->row_fields & (1llu << index)) {
-            /* Need to fetch the value with ecs_field_at() due to being sparse 
-            since this also a reference, we can set the row to 0 as a ref always is a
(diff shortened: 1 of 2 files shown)


Indra-db:
set the row to 0 instead of index, since it's always 0 when it's a ref according to the previous documentaion

Indra-db:
hmm seems like mingw32 test fails, but all other platforms pass 🤔

Indra-db:
setting it back to index also fails, interesting. Not sure what's causing it since it's just on one platform and the logic seems the same to me. 

Nowhere else as far as I can tell is `row_` set.

Indra-db:
I think I've just realized the issue, it should have been sizeof(A) instead of (T) :D. Pure luck it worked on other platforms due to the ptr size being 8 I imagine haha.

Indra-db:
yey tests are passing on all platform now

Indra-db:
Discussed with sander, this won't work as intended behavior and there's a test case missing for this, closing.

PR #1307, opened 2024-08-16
Add 2 sparse queries tests showcasing memory bug to fix (singleton sparse + same term)

Indra-db:
This PR adds 2 test cases, one simple test case, and one more "overloaded" test case to test for other anomalies. 
The test cases are quarantined so CI passes with the intention for you to un-quarantine them upon resolving the bug.

I ran into this bug when converting the Rust code base to v401. This bug currently blocks me from converting / merging to v401.

in short, this bug matches the entity position as singleton on first attempt.

```
    world.component<Position>().add(flecs::Sparse);
    
    world.set<Position>({100, 200});
    
    auto entity = flecs::entity(world)
        .set<Position>({10, 20});
        
q.each([](Position& p, Position& p2) {
        std::cout << "Position: {" << p.x << ", " << p.y << "}" << std::endl;
        std::cout << "Position singleton: {" << p2.x << ", " << p2.y << "}" << std::endl;
});
```

prints

> Position: {100, 200}
> Position singleton: {100, 200}
> Position: {10, 20}
> Position singleton: {100, 200}

### DIFF
--- a/test/cpp/project.json
+++ b/test/cpp/project.json
@@ -836,7 +836,9 @@
                 "named_scoped_rule",
                 "is_valid",
                 "unresolved_by_name",
-                "scope"
+                "scope",
+                "sparse_term_and_term_with_ref",
+                "sparse_multiple_term_and_term_with_ref"
             ]
         }, {
             "id": "SystemBuilder",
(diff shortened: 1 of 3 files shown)


SanderMertens:
The failing test has a mistake in the term indices. When they're changed to

```cpp
        .term_at(1).singleton() // Position
        .term_at(4).singleton() // Mass
        .term_at(6).singleton() // Rotation
```

the test passes.

Indra-db:
just updating the PR, not a bug

image

PR #1308, opened 2024-08-16
fix copy/move-assign operators of World class not freeing existing world

jpeletier:
Fixed bug when overwriting a flecs::World instance with another, either via copy-assign or move-assign. The existing world is not freed, causing a memory leak:

```cpp

flecs::world world1;
flecs::world world2;

world1 = world2;  // leaks, since underlying world_t* in world1 
                  // is not fini'ed before world2's ptr takes over.
                  // same happens if move-assign is used.
```

The PR fixes this issue and adds a test to watch over that.

### DIFF
--- a/distr/flecs.h
+++ b/distr/flecs.h
@@ -20468,6 +20468,7 @@ struct world {
     }
 
     world& operator=(const world& obj) noexcept {
+        release();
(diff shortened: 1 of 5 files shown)


SanderMertens:
Thanks for the PR!

PR #1309, opened 2024-08-16
add world_t pointer to ecs_type_info_t

### LINKED ISSUE #1215: Strings not freed on runtime components
**Describe the bug**

When creating runtime components that contain strings, these strings are not freed on entity destruction, resulting in a memory leak.

**To Reproduce**

This snippet sets up a hook on the `free()` function used by Flecs and prints a message if the pointer to the string we are expecting to be deleted is actually deleted. This deletion never happens, thus there is a memory leak.

```cpp

#include <flecs.h>

ecs_os_api_free_t oldfree;

void *strptr = nullptr; // this will point to the string we are expecting to be freed.

void myfree(void *ptr) {
  if (ptr) {
    if (strptr && strptr == ptr) {
     // print a message if the pointer we are looking for is to be deleted.
      std::cout << "gotcha" << std::endl;
    }
  }
  oldfree(ptr);
}

void main() {

  oldfree = ecs_os_api.free_; // set up a hook on the free() function used by Flecs
  ecs_os_api.free_ = myfree;

  // create a runtime component with one string member
  auto comp = ecs.component("mycomp");
  comp.member(flecs::String, "text");

  flecs::entity e = ecs.entity();
  e.add(comp);

  void *ptr = e.get_mut(comp);

  flecs::cursor cur(ecs, comp, ptr);

  cur.push();
  cur.member("text");
  cur.set_string("hola");
  // get a pointer to the string stored in the component instance:
  uintptr_t *s = (uintptr_t *) cur.get_ptr(); 
  strptr = (void *) (*s);

  cur.pop();

  std::cout << "about to destruct" << std::endl;
  e.destruct(); // at some point here we should expect the string to be destroyed and "gotcha" should be displayed, but it is not.
  std::cout << "after destruct" << std::endl;

}

// output:
// about to destruct
// after destruct
```

I also could not find in the code (`addons/meta`) anything that automatically frees up resources related to runtime components.

**Expected behavior**

When an entity that contains a runtime component is destroyed, all items in that runtime struct should be destroyed.

jpeletier:
As part of preparing to fix #1215, this PR adds a `const world_t* world;` field to `ecs_type_info_t`

This should allow hooks to get `const` access to the world. This is useful for those hooks/destructors, etc, that need to access information about the type they are operating with.


SanderMertens:
Thanks for the PR!

PR #1310, opened 2024-08-16
fix small typos in test running instructions

jpeletier:
Very small typo, but a bit confusing when I was new figuring `bake` out.


SanderMertens:
Ah, that changed in v4. Thanks for the PR!

PR #1312, opened 2024-08-18
Add missing import to c++ stats module

### LINKED ISSUE #1300: Order of module imports affect availability of Units
**Describe the bug**
Importing the 'flecs::stats' module before the 'flecs::units' module results in an error message when trying to register a component member as having a given Unit type.

**To Reproduce**
Consider the following code:
```c++
#include <flecs.h>

int main() {

    flecs::world world;
    world.import<flecs::stats>();
    world.import<flecs::units>();

    struct Foo {
        float x;
    };

    world.component<Foo>().member<flecs::units::length::Meters>("x", &Foo::x);

    return 0;
}
```

Running that code emits the following:
```
error: meta.c: 392: entity 'Meters' for member 'x' is not a unit
```

Swapping the order of those two imports fixes the issue.

NB: This is running on v4.0.0 of flecs.

domdom:
This fixes #1300

### DIFF
--- a/flecs.h
+++ b/flecs.h
@@ -31759,6 +31759,10 @@ inline units::units(flecs::world& world) {
 namespace flecs {
 
 inline stats::stats(flecs::world& world) {
+#ifdef FLECS_UNITS
+    world.import<flecs::units>();
+#endif
+
     /* Import C module  */
     FlecsStatsImport(world);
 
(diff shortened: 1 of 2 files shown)


SanderMertens:
Thanks for the PR!

PR #1313, opened 2024-08-19
Fix some flaky test in meta/ SerializeEntityToJson

jpeletier:
While working on #1215 I noticed some tests of the meta add-on would not pass due to varying entity ids and for alerts, the order of alerts.
So I fixed two tests. For the first one, I gave entities fixed names so we would not be testing against entity ids which can vary.

For the second test, it is more complicated to make it perfect without a JSON library. The order of alerts in the alerts array is of undetermined order, causing the test to fail depending on arbitrary code changes elsewhere, Therefore I wrote an approximation searching for substrings to find the expected alerts in a deterministic way. This should be good enough I think.

### DIFF
--- a/test/meta/src/SerializeEntityToJson.c
+++ b/test/meta/src/SerializeEntityToJson.c
@@ -303,10 +303,14 @@ void SerializeEntityToJson_serialize_component_w_base_w_owned_no_reflection_data
 
     ecs_entity_t base = ecs_new(world);
     ecs_set(world, base, Position, {10, 20});
+    /* Give base a name so we don't have to test unstable entity ids */
+    ecs_set_name(world, base, "Base");
 
     ecs_entity_t e = ecs_new(world);
     ecs_add_pair(world, e, EcsIsA, base);
     ecs_set(world, e, Velocity, {1, 2});
+    /* Give our test entity a name so we don't have to test unstable entity ids */
+    ecs_set_name(world, e, "SomeEntity");
 
     ecs_entity_to_json_desc_t desc = {
         .serialize_inherited = true,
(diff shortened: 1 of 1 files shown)


jpeletier:
regex not available in windows builds. I'll be back...

jpeletier:
fixed!

test/meta/src/SerializeEntityToJson.c
```diff
@@ -986,8 +990,15 @@ void SerializeEntityToJson_serialize_w_2_alerts(void) {
...
+    /* Since alerts can come in any order, test against the two possibilities: */
+    const char* alerts_option1 = "{\"name\":\"e1\", \"alerts\":true, \"components\":{\"flecs.alerts.AlertsActive\":{\"info_count\":0, \"warning_count\":0, \"error_count\":2}, \"Position\":null}, \"alerts\":[{\"alert\":\"position_without_velocity.e1_alert_1\", \"message\":\"e1 has Position but not Velocity\", \"severity\":\"Error\"}, {\"alert\":\"position_without_mass.e1_alert_2\", \"message\":\"e1 has Position but not Mass\", \"severity\":\"Error\"}]}";
+    const char* alerts_option2 = "{\"name\":\"e1\", \"alerts\":true, \"components\":{\"flecs.alerts.AlertsActive\":{\"info_count\":0, \"warning_count\":0, \"error_count\":2}, \"Position\":null}, \"alerts\":[{\"alert\":\"position_without_mass.e1_alert_2\", \"message\":\"e1 has Position but not Mass\", \"severity\":\"Error\"}, {\"alert\":\"position_without_velocity.e1_alert_1\", \"message\":\"e1 has Position but not Velocity\", \"severity\":\"Error\"}]}";
+
+    if(!strcmp(json, alerts_option1) && !strcmp(json, alerts_option2)) {
```
SanderMertens, line 997: Small issue, this should be:
```cpp
if(strcmp(json, alerts_option1) && strcmp(json, alerts_option2)) {
```

jpeletier: whoops, sorry for the oversight -- changed it also to use `ecs_os_strcmp`.

SanderMertens:
Thanks for the PR!

PR #1316, opened 2024-08-20
Have ecs_meta_get_string() work with opaques with .as_type = string

jpeletier:
This PR allows cursors to get strings (`ecs_meta_get_string()`) from opaque types that behave like strings (`.as_type=ecs_string_t`).

This makes it symmetrical, since the opposite operation with `ecs_meta_set_string` which allows to set an opaque's underlying string, was already implemented.


SanderMertens:
Thanks for the PR!

PR #1318, opened 2024-08-20
meta: const string implementation

### LINKED ISSUE #1215: Strings not freed on runtime components

jpeletier:
Note: Includes #1313, so that the tests would pass. It might make sense to review and merge that one first.

This PR implements an opaque type to represent read-only const strings stored in .rodata, so that the automatic destructor I am working on to fix #1215 ignores them, since no destructor is specified for this opaque type.

The PR also maps `build_info_t`'s reflection definition to use these const strings.

In conjunction with #1316, const strings are then more or less transparent from the reflection/cursor point of view. A cursor can read a const string, but won't be able to write to it, since it does not implement `assign_string` overload.

### DIFF
--- a/distr/flecs.c
+++ b/distr/flecs.c
@@ -48537,6 +48537,13 @@ size_t flecs_addon_vec_count(const void *ptr) {
     return flecs_ito(size_t, count);
 }
 
+static
+int flecs_const_str_serialize(const ecs_serializer_t *ser, const void *ptr) {
+    char **data = ECS_CONST_CAST(char**, ptr);
+    ser->value(ser, ecs_id(ecs_string_t), data);
+    return 0;
+}
+
 /* Initialize reflection data for core components */
 static
 void flecs_meta_import_core_definitions(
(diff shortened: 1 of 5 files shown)


SanderMertens:
Thanks for the PR!

PR #1321, opened 2024-08-22
Fix flaky cpp/Enum test

jpeletier:
CPP test `Enum_query_union_enum` was passing depending on compiler / platform due to query returning results in different order.

Fixed by adding all query results to a map and checking correct values deterministically.

Bug affects at least #1318, or whenever a new built-in component is added.


SanderMertens:
Thanks for the PR!

PR #1322, opened 2024-08-24
Add bazel build definitions for examples and add bazel build/test to Ci

reutermj:


SanderMertens:
Awesome, thanks for the PR!

PR #1327, opened 2024-08-29
Properly set and read is_ref in delegate.hpp to fix non-$this variable source components in query iteration

### LINKED ISSUE #1324: Source is not properly taken into account when iterating with the C++ API
**Describe the bug**
When iterating over components with a different source using the C++ API, the new source does not seem to be properly taken into account.

**To Reproduce**
```cpp
flecs::world world;

struct Desc
{
  char const *name;
};

auto e0 = world.entity().set(Desc{"test_0"});
auto e1 = world.entity().set(Desc{"test_1"});

struct Relation
{};

for (int i = 0; i < 10; ++i)
  world.entity().add<Relation>(e0); // Note: we are only referencing e0, so "test_1" should never appear

world.query_builder<const Relation, const Desc>()
  .term_at(0).second("$Instance")
  .term_at(1).src("$Instance")
  .build()
  .each([](Relation const &, Desc const &desc)
  {
    std::cout << desc.name << std::endl;
  });
```

Possible Output:
```
test_0
test_1
<crash or garbage>
```

**Expected behavior**
I would expect "test_0" to be printed 10 times, since 10 entities match, and the component from the other source should be "test_0" for each match.

**Additional context**
It's possible that I'm not using the API properly, but from what I read in the doc and what the flecs explorer suggests as code, this seems to be intended usage.
From other tests I've made, the queries always seem to match the right entities, so it seems it's just the arguments to the callback which are somehow wrong.

Gaspard--:
- set is_ref in populate_self so that it's properly set for any type of query
- always set row to 0 for components with is_ref to avoid indexing garbage

Fixes #1324

### DIFF
--- a/distr/flecs.h
+++ b/distr/flecs.h
@@ -25659,6 +25659,7 @@ private:
     void populate_self(const ecs_iter_t *iter, size_t index, T, Targs... comps) {
         fields_[index].ptr = ecs_field_w_size(iter, sizeof(A), 
             static_cast<int8_t>(index));
+        fields_[index].is_ref = iter->sources[index] != 0;
         populate_self(iter, index + 1, comps ...);
     }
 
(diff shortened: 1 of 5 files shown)


SanderMertens:
Nice! Can you:
- Add a test case that reproduces the bug
- Run bake once on the repository so the distr/flecs.h files get updated? (this is what's causing CI to fail)

Gaspard--:
I tried to cover some range of scenarios in the test. Didn't add optional arguments because I felt like the test was getting a bit long.

SanderMertens:
Thanks, almost there:
- Can you split up the test into three individual test cases?
- You'll need to add the test cases to the `test/cpp/project.json` file for them to be executed

Gaspard--:
I thought about is some more and I'm pretty sure these 2 tests cover all cases. I wanted to make sure reading / writing relationships was consistent originally, but I'm pretty sure that superfluous.

SanderMertens:
Thanks for the PR!

PR #1336, opened 2024-09-05
Add perf trace when sleeping due to target_fps

waywardmonkeys:


SanderMertens:
Nice, thanks for the PR!

PR #1352, opened 2024-09-12
Add flecs::iter::other_table()

SanderMertens:
Summary: This diff upstreams a change that adds an `flecs::iter::other_table()` method.

Differential Revision: D62449718


Indra-db (changes requested):
would like to see some comments on what other_table is, should be used for

PR #1354, opened 2024-09-13
fix leak in ecs_clone

jpeletier:
When working on #1330 I noticed that when copying an entity with `ecs_clone(...,true)`, a component's ctor was being called twice, the second time over already initialized memory. This causes a memory leak of components that allocate dynamic memory in the constructor.

Even if the component is trivial, the registered constructor was being called twice during a clone anyway. This PR fixes that and adds a test to make sure there are no leaks.

The bug does not affect copies created while instantiating a prefab.


SanderMertens:
Thanks for the PR!

PR #1356, opened 2024-09-16
fixed copy-paste error in rtt tests

jpeletier:
When examining the diff again I saw a copy-paste error where I had overlooked replacing "array" with "vector". This fixes that. My apologies.


SanderMertens:
Thanks for the PR!

PR #1364, opened 2024-09-19
Add C++20 module build target

### LINKED ISSUE #1317: C++20 module support
**Describe the problem you are trying to solve.**

`flecs.h` isn't small, and can slow down compilation of a file about 1-2 seconds on my machine.
This can add up over multiple files.
C++ modules allow to reduce this by only exposing relevant symbols and pre-compiling the module so that it can be reused.
I tried wrapping `flecs.h` into a module manually, but it seems due to how `new` is used within the library this does not work.
Maybe I'm missing something though.

**Describe the solution you'd like**

Add C++20 module support directly into flecs, allowing something like `import flecs;`.

From what I checked in [flux](https://github.com/tcbrindle/flux), C++ module support boils down to adding a define (something like `FLECS_EXPORT`) which expands to `export` or nothing depending on another define.
Then, a module file is provided by the library which exports a module, enables this define and then includes the header to export the symbols.

Gaspard--:
Usage in client cmake:
```cmake
set(FLECS_M0DULE ON CACHE BOOL "Build flecs module" FORCE)
add_subdirectory(path/to/flecs)
# If the c++ version of the project isn't c++20, the matching version must be set manually
target_compile_features(flecs-module PUBLIC cxx_std_23)
target_link_libraries(
  someProjectName
  PRIVATE
  flecs-module
)
```

In the code, it is then possible to use `import flecs;` to import all the flecs symbols.

~Still investigating why reflection seems to break, disabled in the module for now.~ Fixed.

Maintainance:
- structs exposed by flecs must be tagged with `FLECS_API_STRUCT`
- code within the c++ addon tagged with `FLECS_API` are exposed
- globals must be tagged with `FLECS_API_GLOBAL`, which becomes `static` in headers, and export in the module.

Closes #1317

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,6 +7,7 @@ option(FLECS_STATIC "Build static flecs lib" ON)
 option(FLECS_SHARED "Build shared flecs lib" ON)
 option(FLECS_PIC "Compile static flecs lib with position independent code (PIC)" ON)
 option(FLECS_TESTS "Build flecs tests" OFF)
+option(FLECS_M0DULE "Build flecs C++20 module" OFF)
 
 include(cmake/target_default_compile_warnings.cmake)
 include(cmake/target_default_compile_options.cmake)
@@ -73,6 +74,10 @@ endif()
 if(FLECS_TESTS)
     enable_testing()
     add_subdirectory(test)
+endif()
+
+if(FLECS_M0DULE)
+    add_subdirectory(module)
 endif()
 
 message(STATUS "Targets: ${FLECS_TARGETS}")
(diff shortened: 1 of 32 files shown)


Gaspard--:
This is a first working version, I don't know flecs enough for what approaches might be preferred for the define names and usage.

Gaspard--:
Ah, seems my fix for the modularised version of `flecs` on windows broke the normal build.
I guess this means we need another define for the specific case of `operator new`.
Currently testing gcc, clang++ and msvc work with the modularized version.

Gaspard--:
It seems it's rather complicated to make g++ work with the module version, and since g++'s support of modules is incomplete, I don't think it's worth it to support it right now.

Gaspard--:
g++'s module support is incomplete, and the work around seems to big to be worth it right now.

Gaspard--:
@SanderMertens can I have your thoughts on this approach ?
Also, if this is not mergeable, I could just maintain the fork until C++ modules mature a little more.

Gaspard--:
Closing this PR since #1410 supersedes it.

PR #1374, opened 2024-09-26
Added ecs_table_clear_entities

gsnook:
Added ecs_table_clear_entities to allow users to clear a table's entities while optionally retaining its memory allocations. This can help reduce allocation churn when overwriting table entities using ecs_bulk_init. 

Discord Conversation: https://discord.com/channels/633826290415435777/1090412095893422101/1288146063982858242

### DIFF
--- a/distr/flecs.c
+++ b/distr/flecs.c
@@ -37034,14 +37034,12 @@ const ecs_entity_t* ecs_table_entities(
     return table->data.entities;
 }
 
-/* Cleanup, optionally run OnRemove, clear entity index, deactivate table, optionally free allocations */
+/* Cleanup, no OnRemove, delete from entity index, deactivate table, retain allocations */
 void ecs_table_clear_entities(
     ecs_world_t* world,
-    ecs_table_t* table,
-    bool notify,
-    bool deallocate)
+    ecs_table_t* table)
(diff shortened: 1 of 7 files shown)


SanderMertens:
Looks good! Can you add test cases for the following scenarios:
 - clear an empty table
 - clear a table with >1 entities and verify they're no longer alive afterwards
 - clear a table with >1 entities, repopulate the table & make sure the new entities are matched by a query
 - clear a table with >1 entities and verify that on_remove hooks are invoked
 - clear a table with >1 entities and verify that OnRemove observers are invoked

gsnook:
Tests added. Undecided on whether they should all live in Table.c or move the observer and hook tests to Observer.c/ComponentLifecycle.c.  Any preference?

SanderMertens:
Tests are good in Table.c

Are the dealloc/notify parameters necessary? I know those exist in the internal APIs , but I can't think of scenarios where you'd want that in the public API.

gsnook:
Probably not. I added them for additional flexibility, but it wouldn't be a bad idea to remove them and avoid confusion. Someone can always propose a PR with them if a need arises. I'll post an update with the parameters removed.

PR #1383, opened 2024-09-30
Add FLECS_NO_ALERTS support

pfeodrippe:
I was having some runtime crash with FLECS_REST and FLECS_TIMER (the last one which the alerts module uses) while compiling flecs with zig, but wasn't unable to disable FLECS_ALERTS, so I'm assuming that's missing unintentionally and I'm adding it here \o

### DIFF
--- a/distr/flecs.h
+++ b/distr/flecs.h
@@ -10314,6 +10314,9 @@ int ecs_value_move_ctor(
 #ifdef FLECS_NO_SYSTEM
 #undef FLECS_SYSTEM
 #endif
+#ifdef FLECS_NO_ALERTS
+#undef FLECS_ALERTS
+#endif
 #ifdef FLECS_NO_PIPELINE
 #undef FLECS_PIPELINE
 #endif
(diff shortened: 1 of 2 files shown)


BeanCheeseBurrito:
Are you running into this issue in the picture below? I had a similar issue with the FLECS_TIMER addon that was caused by zig compiling sanitized builds by default. The fix for me was to compile flecs with `-fno-sanitize=undefined`.
```zig
lib.addCSourceFile(.{ .file = b.path("flecs.c"), .flags = &.{"-fno-sanitize=undefined"} });
```

BeanCheeseBurrito:
> @BeanCheeseBurrito Yes, same thing here, will try that. How did you found that? Using zig tooling or something else?

I used gdb. The illegal instruction error is something I've gotten in the past when compiling other projects with zig. https://github.com/ziglang/zig/wiki/FAQ#why-do-i-get-illegal-instruction-when-using-with-zig-cc-to-build-c-code

SanderMertens:
Thanks for the PR!

PR #1385, opened 2024-10-02
remove FLECS_API in block_allocator.c -> fixes dll import issue

Indra-db:
fixes error: flecs.c(28687): Error C2491 : 'flecs_bfree_w_dbg_info': definition of dllimport function not allowed


SanderMertens:
Thanks for the PR!

PR #1397, opened 2024-10-13
Add support for type checked term_at<T> and term_at<T>(i)

deathbeam:
First method is simple loop over terms to find term for T. Its slower than using index directly, but sometimes type safety is more preferred and when the cost is only on initialization.

Second method is same as term_at(i) but with type checked assert. For middle ground to still keep the safety but also performance.

This is mostly to mirror the C# binding PR where this is slightly more useful but maybe cpp would benefit too:

https://github.com/BeanCheeseBurrito/Flecs.NET/pull/52

Also disclaimer, I dont rly work with cpp often so I might've easily missed something, but I made best effort to write some tests and at least run them and see if they pass :d

Example use:

Current method

```cpp
world.System<Position, Velocity>()
  .term_at(1)
  .singleton();
```

Type checked (throws assert if term_at(1) isnt Velocity type)

```cpp
world.System<Position, Velocity>()
  .term_at<Velocity>(1)
  .singleton();
```

Type checked + resolved (throws assert if there is no Velocity component):

```cpp
world.System<Position, Velocity>()
  .term_at<Velocity>()
  .singleton();
```


copygirl:
Would it make sense to have the parameter-less `term_at` named differently?

deathbeam:
Maybe just .term<>() and then also add .term(flecs::entity_t component) to match other apis?

deathbeam:
Renamed term_at<T> to term<T>. Also updated the logic to account for pairs properly

deathbeam:
Looks like I cant use term<T> as it fails on build here so changing back to term_at I guess:

https://github.com/SanderMertens/flecs/actions/runs/11327040474/job/31497292704?pr=1397

SanderMertens:
LGTM! I'll merge when the conflicts are resolved.

deathbeam:
Completely forgot I had this PR, rebased, thanks for reviewing

SanderMertens:
Thanks for the PR!

Issue #1386, opened 2024-10-03
Can't compile if I include magic_enum using MSVC.


xagalvarez:
**Describe the bug**

If I include magic_enum header before flecs header I get an error with flecs, if I switch it I get an error with both headers.

`C:\Users\Agustin\Desktop\Chotardum Online\Aeldur\.idea\cmake-debug\_deps\external_flecs-src\include\flecs\addons\cpp\utils\enum.hpp(337): warning C4003: not enough arguments for function-like macro invocation 'min'
C:\Users\Agustin\Desktop\Chotardum Online\Aeldur\.idea\cmake-debug\_deps\external_flecs-src\include\flecs\addons\cpp\utils\enum.hpp(335): error C2589: '(': illegal token on right side of '::'
C:\Users\Agustin\Desktop\Chotardum Online\Aeldur\.idea\cmake-debug\_deps\external_flecs-src\include\flecs\addons\cpp\utils\enum.hpp(319): note: This diagnostic occurred in the compiler generated function 'std::underlying_type<_Ty>::type flecs::_::enum_type<E>::reflection_init::handle_constant(std::underlying_type<_Ty>::type,flecs::world_t *)'
C:\Users\Agustin\Desktop\Chotardum Online\Aeldur\.idea\cmake-debug\_deps\external_flecs-src\include\flecs\addons\cpp\utils\enum.hpp(335): note: the template instantiation context (the oldest one first) is
C:\Users\Agustin\Desktop\Chotardum Online\Aeldur\.idea\cmake-debug\_deps\external_flecs-src\include\flecs\addons\cpp\utils\enum.hpp(301): note: while compiling class template 'flecs::_::enum_type'
C:\Users\Agustin\Desktop\Chotardum Online\Aeldur\.idea\cmake-debug\_deps\external_flecs-src\include\flecs\addons\cpp\utils\enum.hpp(335): error C3878: syntax error: unexpected token '(' following 'expression'
C:\Users\Agustin\Desktop\Chotardum Online\Aeldur\.idea\cmake-debug\_deps\external_flecs-src\include\flecs\addons\cpp\utils\enum.hpp(319): note: This diagnostic occurred in the compiler generated function 'std::underlying_type<_Ty>::type flecs::_::enum_type<E>::reflection_init::handle_constant(std::underlying_type<_Ty>::type,flecs::world_t *)'
[...]

PR #1402, opened 2024-10-16
Fix conflicting identifier between flecs and magic_enum

SanderMertens:
Summary: Fix conflicting identifier between flecs and magic_enum that causes compilation to fail.

Differential Revision: D64483027


SanderMertens:
Fixes #1386

PR #1405, opened 2024-10-17
[meta] Getter interface for opaque collection types

jpeletier:
As a continuation to #1403, this PR adds getter opaque support for collection types.
One can now differentiate between reading and writing operations with a cursor, by means of setting a `get_member` or `get_element` callback:

```cpp
  struct MyOpaque {
    int x;
    bool x_has_value = false;
    int y;
    bool y_has_value = false;
  };

  auto my_component = ecs.component<MyOpaque>();
  my_component.opaque(ecs.component().member<int>("a").member<int>("b"))
      .ensure_member([](MyOpaque *data, const char *member) -> void * {
        if (!strcmp(member, "a")) {
          data->x_has_value = true;
          return &data->x;
        } else if (!strcmp(member, "b")) {
          data->y_has_value = true;
          return &data->y;
        }
        return nullptr;
      })
      .get_member([](const MyOpaque *data, const char *member) -> const void * {
        if (!strcmp(member, "a") && data->x_has_value) {
          return &data->x;
        } else if (!strcmp(member, "b") && data->y_has_value) {
          return &data->y;
        }
        return nullptr;
      })
      .serialize([](const flecs::serializer *s, const MyOpaque *data) {
        if (data->x_has_value) {
          s->member("a");
          s->value(data->x);
        }
        if (data->y_has_value) {
          s->member("b");
          s->value(data->y);
        }

        return 0;
      });

  MyOpaque instance;

  flecs::cursor cur(ecs, my_component, &instance);

  cur.push();
  cur.member("a");
  cur.set_int(79);
  std::cout << "a = " << cur.get_int() << std::endl;  // value = 79
  cur.member("b");
  if (cur.get_read_ptr()) {
    std::cout << "b = " << cur.get_int() << std::endl;
  } else {
    std::cout << "b has no value" << std::endl;
  }

  std::cout << ecs.to_json(&instance) << std::endl;
  // output:
  // a = 79
  // b has no value
  // {"a":79}
```

With collections, one can now do the following. The below implements a pseudo sparse array opaque that is disguised to Flecs as a vector of floats:

```cpp
  using MyMap = std::map<size_t, float>;

  auto float_vector = ecs.vector(flecs::F32);

  auto my_component = ecs.component<MyMap>();
  my_component.opaque(float_vector)
      .count([](const MyMap *m) -> size_t { return 100; })
      .resize([](MyMap *m, size_t size) {})
      .ensure_element([](MyMap *m, size_t elem) -> void * { return &(*m)[elem]; })
[...]

PR #1409, opened 2024-10-21
[meta] Automatic opaque serialization

jpeletier:
**Note: must merge #1405 first**

This PR builds on the previous two to add automatic opaque serialization for those opaques defining the appropriate `get_*` callbacks.

When defining serialization for an opaque, you have to define a `serialize` callback. If this callback is not defined, serialization will fail.

With this change, when a `serialize` callback is not defined, the `meta` module will try to find a way to serialize the opaque if the opaque has defined another callback that can be used.

This is useful if you have already defined a `get_` callback that does the work and don't want to define a serializer.

If you do define a serializer, the serializer takes precedence and it is used as before.

For example:

```cpp
  struct IntOpaque {
    int value;
  };

// BEFORE:
  ecs.component<IntOpaque>()
      .opaque(flecs::I32)
      .serialize([](const flecs::serializer *ser, const IntOpaque *data) -> int { 
        ser->value(data->value);
        return 0; 
    });

// Now: 
  ecs.component<IntOpaque>()
    .opaque(flecs::I32)
    .get_int([](const IntOpaque *data) -> int64_t { 
      return data->value; 
    });
```

It is still recommended to write a serializer if what you are doing with the type is to fully serialize it to JSON or something similar. 

If, however, you are using reflection to access portions of opaque data randomly, the serializer could be slower/inefficient because it dumps the whole content, whereas the `get_` callbacks provide the opaque with a way to go straight to the requested data.

For example, the following will now work without a serializer:

```cpp
  using IntVector = std::vector<int>;

  ecs.component<IntVector>()
      .opaque(ecs.vector<int>())
      .get_element([](const IntVector *src, size_t i) -> const void * { return src->data() + i; })
      .count([](const IntVector *src) -> size_t { return src->size(); });

  IntVector vec = {1, 2, 3};

  // work with IntVector accessing a random element:
  flecs::cursor cur(ecs, ecs.component<IntVector>(), &vec);
  cur.push();
  cur.elem(2);
  std::cout << "vec[2] = " 
            << cur.get_int()  // will call opaque's get_element()
            << std::endl;

  // Can also serialize the whole vector to JSON, without having to write a serializer,
  // at a small performance cost:
  std::cout << "JSON: " << ecs.to_json(&vec) << std::endl;

  // output:
  // vec[2] = 3
  // JSON: [1, 2, 3]
```


SanderMertens:
Similar to my comment on the other PR- I'm not sure how this works if you have a type with multiple members. I don't think the edge case of a type with a single member warrants extending the interface.

SanderMertens:
> With this change, when a serialize callback is not defined, the meta module will try to find a way to serialize the opaque if the opaque has defined another callback that can be used.

This is the opposite of what should happen. A `serialize` callback should _always_ be implemented that serialized from opaque type to as_type. When a conversion is required to another type, it should reuse the existing conversion code to go from opaque value -> serialize -> as_type value -> convert to destination type.

Closing as this is not the right way to go about this.

PR #1411, opened 2024-10-23
Improve macOS CI

waywardmonkeys:
* Use same capitalization as GitHub: `macOS-*` is now `macos-*`
* `build-cmake`: Now depends on `build-macos` in addition to `build-linux` so that a build failure in `build-macos` will stop this step from running, just like already happens for Linux.
* `build-cmake`, `test-c-unix`, `test-cmake`: Run on `macos-13` and `macos-latest`. This was running on `macos-14` and `macos-latest` which is from when `macos-latest` was still `macos-13`. We keep `macos-13` for now since this is still an Intel build while `macos-14` and later are Apple Silicon hardware.
* `test-cpp-macos`: Remove `macos-12` as this will be turned off in the coming months at GitHub. Add `macos-15` which is in public preview. Update Xcode versions to the latest for each major version and run some of the older ones on `macos-13` now that `macos-12` is gone.

Information about the pending removal of `macos-12` can be found here: https://github.com/actions/runner-images/issues/10721

Information about Xcode releases can be found here: https://xcodereleases.com


SanderMertens:
Thanks for the PR!

PR #1414, opened 2024-10-26
[meta] Notify on enum modifications

jpeletier:
This PR makes sure that when an enum changes, observers on `EcsEnum` (`flecs::Enum`) are notified.

This makes the api coherent with structs, that is, when a member is added to a struct, observers watching `EcsStruct` (`flecs::Struct`) are notified.

Normally, structs/enums/etc are only modified during startup logic as they are constructed. These notifications help serialization libraries synchronize what they know about the type.


SanderMertens:
Thanks for the PR!

PR #1416, opened 2024-10-29
Remove reinterpret_casts when assigning `ecs_ctx_free_t` which were breaking aliasing rules

Gaspard--:
The code was casting functions pointers of the signature `void (*)(T *)` to `void (*)(void *)` and then calling them on a `void *`. This breaks aliasing rules and is undefined behaviour. This commit removes these casts by changing the template to take a `void *` directly.

This was caught by `-fsanitize=undefined` under clang.
I'm suspicious of a couple other places which seem to `reinterpret_cast` function pointers, but since I have no concrete evidence I'd rather not touch them.

### DIFF
--- a/distr/flecs.h
+++ b/distr/flecs.h
@@ -16905,10 +16905,10 @@ namespace _
 struct placement_new_tag_t{};
 constexpr placement_new_tag_t placement_new_tag{};
 template<class Ty> inline void destruct_obj(Ty* _ptr) { _ptr->~Ty(); }
-template<class Ty> inline void free_obj(Ty* _ptr) { 
+template<class Ty> inline void free_obj(void* _ptr) {
     if (_ptr) {
-        destruct_obj(_ptr); 
-        ecs_os_free(_ptr); 
+        destruct_obj(static_cast<Ty *>(_ptr));
+        ecs_os_free(_ptr);
     }
 }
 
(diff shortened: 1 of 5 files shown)


SanderMertens:
> The fix seems more complicated than https://github.com/SanderMertens/flecs/pull/1416

This comment in #1417 suggests that this is not a correct fix. Should this PR be closed?

Gaspard--:
Sorry if I was unclear. In this case, I changed the function signature of the function we assign to the callback itself to `void (*)(void *)`, so that we don't need a `reinterpret_cast` at all. This is possible because we provide the callback in flecs itself.

In the case of the linked issue, the user provides the callback, so it's not possible to change the signature trivially without harming user experience.

jpeletier:
I would merge this anyway. It does not solve the problem for the opaques c++ API in #1417 but this PR is valuable because it does clean up the code by fixing `free_obj()`'s signature and now there is much less clutter.

Regarding #1417 we can probably get away with undefined behavior. That cast, while undefined behavior by the standard, works in practicality, unless you are in some very old or obscure embedded architecture. The solution is much worse: there is no other option but to pass the function pointer along and make a double call, and we'd need a place for it, etc.

SanderMertens:
Just merged another PR that also fixed this issue! Sorry- should have merged this one since it was older, but got confused by the comment in the issue 🫠

PR #1420, opened 2024-10-31
Remove unnecessary version attribute from flecs bazel module definition

reutermj:
The version attribute is not required in bazel module definitions. Currently, the only way to depend on flecs with bzlmod is through `git_override`, which overrides the version with a specific commit anyways. if flecs becomes distributed on Bazel central, this decision will need to be reevaluated, but for now, it's easier to leave it off instead of having to maintain keeping the module version up to date with the current release.

Also updated the docs with guidance on how to configure renovate to autoupdate flecs dependency.


SanderMertens:
Thanks for the PR!

PR #1435, opened 2024-11-13
Add assertion to prevent max_id overflowing

Ukendio:
The entity_index only allows up to 2^32 indexes, even though overflowing would be very unlikely with recycling it is still a possibility and would lead to hard-to-find bugs. This PR aims to make the assertion to prevent this. 

P.S. pls squash commit it because I made a mistake twice.


src/storage/entity_index.c
```diff
@@ -286,8 +289,11 @@ uint64_t* flecs_entity_index_new_ids(
     for (i = 0; i < to_add; i ++) {
         uint32_t id = (uint32_t)++ index->max_id;
 
+        ecs_assert(index->max_id <= UINT32_MAX, ECS_INVALID_OPERATION,
```
SanderMertens, line 292: Can you add a test that also triggers this assert?

PR #1436, opened 2024-11-13
Fixed UBSan runtime error: calling function through pointer to incorrect type

jspohr:
UBSan error was:
> call to function void flecs::_::free_obj<flecs::_::each_delegate<>> through pointer to incorrect function type 'void (*)(void *)'

It's caused by reinterpret_casting the free_obj template to ecs_ctx_free_t and then calling it through that function type. Changing the free_obj signature fixes it. However, I don't know if you're planning on checking UBSan, so this might not be something you care about. IMHO it improves readability to get rid of the reinterpret_casts, even though you lose type safety when calling free_obj  directly.


SanderMertens:
LGTM, thanks for the PR!

PR #1456, opened 2024-11-30
WIP: Visitor pattern implementation for serialization

garrett-is-a-swann:
Reworks/merges the JSON/Script serialization functions to use a single entity-component "graph" traversal mechanism. This mechanism can also be used by user code in order to perform arbitrary serializations, and/or as a map/reduce mechanism.


distr/flecs.h
```diff
@@ -15331,6 +15331,33 @@ typedef struct ecs_serializer_t {
...
+ */
+typedef struct ecs_visitor_desc_t {
+    /** Here Be Visitor Functions
+     * If the visit has a return, it has some form of flow control.
+     */
+
+    /** Visit Primitives */
+    void (*visit_f32)(const ecs_f32_t value, void* user_data);
+    void (*visit_f64)(const ecs_f64_t value, void* user_data);
+
```
copygirl, line 15344: This not appear to have all the same functions that appear in `include/flecs/addons/meta.h`?

garrett-is-a-swann: Those are functions for opaque type serialization. Currently all that is uploaded is an initial (wip) commit since I'm still working on a commit after raising the abstraction all the way to `ecs_array_to_json` and similar serialization entry points. 

To get the gist, you can see https://github.com/SanderMertens/flecs/pull/1456/files#diff-183bddbbb7fb95a645f2bf0b20f055456269a7d4994f37e63a60844b4042ca12R155. The struct you quoted is passed into the serialization functions for each type you want to use. It will call into those opaque serialization functions you mention.

SanderMertens:
I like the idea of a visitor that understands the reflection data and provides a simpler user-facing API. The current reflection API is complex, and is also not documented very well.

However, I think such a visitor API would be a better fit for C++ as there the compiler is able to optimize out virtual calls if it knows the type of the visitor. In C having this many callbacks hurts performance since they can't be inlined. Both the JSON serializer and flecs script are performance critical, so I don't think this PR is moving things in the right direction.

Some thoughts on improving the usability of the reflection API:
- A reflection manual is sorely needed
- Examples that show how to iterate the `TypeSerializer`
- I'd like to explore whether the `flecs::cursor` API can be extended to provide easier iteration of members/elements
- A C++ visitor-based approach could start out as an external repository
- If extending existing APIs proves too cumbersome, the visitor could get upstreamed to flecs

SanderMertens:
Closing as this is not intended to get merged.

PR #1457, opened 2024-11-30
Update Systems.md

iHR4K:
changes for world name


iHR4K:
I'm not sure which one, but they have a discrepancy in one scope. Another scope later in the file uses the ecs name. 
Since I'm here - I can do one or another, please let me know what you think

SanderMertens:
I think `world` is more clear.

iHR4K:
awesome :)

PR #1462, opened 2024-12-09
Move threading check in HTTP server to start/stop methods.

basicer:
When using the flecs_explorer_request hook from wasm with FLECS_NO_OS_API_IMPL set, the lack of threading support prevents the http server from being created. We don't need lock support if we don't plan to actually launch the resulting server, so move the check that we have that support to starting/stopping the server.


SanderMertens:
Thanks for the PR!

PR #1481, opened 2024-12-19
Custom underlying enum type

### LINKED ISSUE #696: Add support for enums with custom integer types
**Describe the bug**
Unsure if this is a bug or a feature request:

The following snippet compiles and triggers a heap corruption runtime check on MSVC
```c++
enum class ETest : uint8_t
{
    A, B
};
int main()
{
    
    flecs::world ecs;

    ecs.component<ETest>();
    ecs.entity().set<ETest>(1);
    return 0;
}
```

**Expected behavior**
I don't think the above should be allowed to compile?

SanderMertens:
Fixes #696

### DIFF
--- a/distr/flecs.h
+++ b/distr/flecs.h
@@ -15661,11 +15661,11 @@ typedef struct ecs_enum_constant_t {
     /** May be set when used with ecs_enum_desc_t */
     int64_t value;
 
-    /** Should not be set by ecs_enum_desc_t */
-    ecs_entity_t constant;
-
     /** For when the underlying type is unsigned */
     uint64_t value_unsigned;
+
+    /** Should not be set by ecs_enum_desc_t */
+    ecs_entity_t constant;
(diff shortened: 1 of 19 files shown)


distr/flecs.h
```diff
@@ -15659,14 +15659,19 @@ typedef struct ecs_enum_constant_t {
     const char *name;
 
     /** May be set when used with ecs_enum_desc_t */
-    int32_t value;
+    int64_t value;
+
+    /** For when the underlying type is unsigned */
+    uint64_t value_unsigned;
```
jpeletier, line 15665: Why not a union?

SanderMertens: This type is used in `ecs_enum_desc_t`/`ecs_bitmask_desc_t` so it needs to be easy to assign. If it was turned into a union, you would have to set two fields, one to indicate which branch is used, and one to set the actual value.

PR #1493, opened 2024-12-30
Update Term flags to constexpr for C++

Reddy-dev:


Reddy-dev:
Will add a FLECS_CONSTEXPR macro

Reddy-dev:
would then support amalgamated as well

Reddy-dev:
can also be future proofed for C23

SanderMertens:
I'm closing this PR as per the discussion on discord- `constexpr` cannot be consistently applied to all constants since many are defined in C source files. Since it doesn't provide a functional benefit, I prefer having a single consistent way to declare constants.

PR #1494, opened 2025-01-03
Fix possible handle leak

peter15914:
Calls to ```fclose()``` were missing in some cases when ```goto error``` was used.
Therefore, I added ```fclose()```.
The check ```if (file)``` is needed, because ```goto error``` is also used when file was not successfully opened.

### DIFF
--- a/distr/flecs.c
+++ b/distr/flecs.c
@@ -12997,6 +12997,9 @@ char* flecs_load_from_file(
 
     return content;
 error:
+    if (file) {
+        fclose(file);
+    }
     ecs_os_free(content);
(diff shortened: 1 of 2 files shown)


SanderMertens:
LGTM, can you also update the flecs.c file in the `distr` folder? That's what is causing CI to fail.

peter15914:
> LGTM, can you also update the flecs.c file in the `distr` folder? That's what is causing CI to fail.

Thanks for the review!
I updated flecs.c.

SanderMertens:
Thanks for the PR!

PR #1506, opened 2025-01-07
Fix each_ref_field not casting to int32_t for ecs_field_at_w_size's ow parameter

Megafunk:
This is used in the C++ API for iterating sparse component data and appears to cast to int8 instead of the intended int32.

Repro with making 2000 sparse components and iterating them in a system:
```cc
    flecs::world ecs;

    class Vec4Sparse {
    public:
        float F32[4];
    };

    ecs.component<Vec4Sparse>().add(flecs::Sparse);

    for (int i = 0; i < 2000; ++i) {
        auto e = ecs.entity().set<Vec4Sparse>({});
        std::cout << " Made entity id: " << e.id() << "\n";
    }

    ecs.system<Vec4Sparse>().each([](flecs::entity e, Vec4Sparse& p) {
        // crash happens in .each field templates
        std::cout << e.id() << "Sparse data: {" << p.F32[0] << ", " << p.F32[1] << "}\n";
    });

    ecs.progress();
```

### DIFF
--- a/distr/flecs.h
+++ b/distr/flecs.h
@@ -26543,7 +26543,7 @@ struct each_ref_field : public each_field<T> {
 
         if (field.is_row) {
             field.ptr = ecs_field_at_w_size(iter, sizeof(T), field.index, 
-                static_cast<int8_t>(row));
+                static_cast<int32_t>(row));
         }
     }
 };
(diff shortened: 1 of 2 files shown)


Megafunk:
Argh, I forgot to bake! Hold on

SanderMertens:
Thanks for the PR!

PR #1514, opened 2025-01-11
Add missing addons/app.c to meson.build

justinian:
The App addon was missing from the meson build.


SanderMertens:
Thanks for the PR!

PR #1521, opened 2025-01-16
Fix query copy operations with default init queries

Thinkofname:
Copy ctor/assign for C++ queries would abort in `flecs_poly_claim` if called with a default initialized query. This fixes that and `flecs_poly_release` not being called for the query being copied over during assignment.


SanderMertens:
Thanks for the PR!

PR #1541, opened 2025-01-30
fix(musl): call srand in ecs_os_api setup to prevent `rand()` returning 0

Indra-db:
on first call where libc implementation on different platforms (like musl) might differ how rand() behaves.

fixes https://github.com/Indra-db/Flecs-Rust/issues/222 as well.


SanderMertens:
I don't want to call `srand` in Flecs, since this modifies global state. Applications might call `srand` themselves with a specific value at an unknown time (before or after flecs is initialized), and this would overwrite that.

PR #1551, opened 2025-02-05
Add Sep option for flecs::entity constructor (in C++ API)

Reddy-dev:


SanderMertens:
LGTM! Can you add a test in the `cpp/Entity` suite to make sure this works as expected?

Reddy-dev:
Should work now

SanderMertens:
Awesome thank you!

PR #1560, opened 2025-02-10
Fixing Enum ambiguous calls

Reddy-dev:


Reddy-dev:
Now should be added

SanderMertens:
Can you add a test that fails without this fix?

Reddy-dev:
> Can you add a test that fails without this fix?

the issue is a compiler error though, i can show you what code errored?

Reddy-dev:
`TestEntity.add<ETestEnum>(ETestEnum::One);`

SanderMertens:
LGTM!

PR #1568, opened 2025-02-13
Fix members named `_` not being created in structs

GsLogiMaker:
This PR fixes an issue I found with `ecs_struct_init`. When attempting to initialize a struct with a member named `_` the member would silently not be created at all. The issue was that the method used to create the members, `ecs_entity`, would return the *any* entity instead of creating a new one. Swapping that method for `ecs_new_from_path` fixes the issue.

### DIFF
--- a/distr/flecs.c
+++ b/distr/flecs.c
@@ -47167,9 +47167,7 @@ ecs_entity_t ecs_struct_init(
             goto error;
         }
 
-        ecs_entity_t m = ecs_entity(world, {
-            .name = m_desc->name
-        });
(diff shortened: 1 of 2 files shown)


SanderMertens:
Looks good to me! After you update distr/flecs.c as well CI should pass.

GsLogiMaker:
Cool! I updated it manually this time, but for future reference, is there a way to update it automaticly from `src/`?

SanderMertens:
It will automatically update when you run `bake` in the root of the repository.

PR #1576, opened 2025-02-15
fix(prefab): prevent infinite recursion on self-referencing disabled prefabs

Indra-db:
When a prefab type was added as a component of itself and subsequently disabled, the enabling/disabling code entered an infinite recursion. This commit introduces a check to detect self-references in disabled prefabs.


Indra-db:
my repo seems broken, I never submitted component.hpp, sigh

Indra-db:
will submit new PR

Indra-db:
seems like bake update was the cause of it, it put my repo back in time

PR #1581, opened 2025-02-17
Use correct world in term_builder_i::singleton(), Add missing enum tests

BeanCheeseBurrito:
When calling ``term_builder_i::singleton()``, there's a typo where the term's source id isn't being set correctly for pairs (always 0) because it is creating and passing a new world  to ``ecs_get_alive`` instead of using the world that is returned by ``world_v()``. 
https://github.com/SanderMertens/flecs/blob/98d13a6907996bfc4c9faacddfc8b102a0b9af64/include/flecs/addons/cpp/mixins/term/builder_i.hpp#L404

Bake generated some missing enum tests so I included those as well.

### DIFF
--- a/distr/flecs.h
+++ b/distr/flecs.h
@@ -29993,7 +29993,7 @@ struct term_builder_i : term_ref_builder_i<Base> {
         if (!ECS_IS_PAIR(sid)) {
             term_->src.id = sid;
         } else {
-            term_->src.id = ecs_pair_first(world(), sid);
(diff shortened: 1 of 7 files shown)


SanderMertens:
LGTM! Can you add a test for the failing behavior?

BeanCheeseBurrito:
Added a test case.

SanderMertens:
LGTM!

PR #1582, opened 2025-02-18
fix(docs): cascade queries need to be cached.

Indra-db:


SanderMertens:
As of commit 2363674e8f44a4ac4f4676195174853c62046335 queries with a cascade term will automatically turn into cached queries.

PR #1587, opened 2025-02-23
Use Low Ids for Untyped Components

Reddy-dev:


Reddy-dev:
Not sure if this is the API you want or not though

SanderMertens:
LGTM, can you add tests for the new functions?

Reddy-dev:
Should be added now, added 1 basic test for testing if the id is less than component high id, not sure if you'd want more test if you can think of any.

SanderMertens:
Thanks for the test!

PR #1596, opened 2025-03-08
Add missing Ref tests to project.json

jpeletier:
In reviewing code, I just realized two `Refs` tests exist but are not being run because they are missing from the corresponding `project.json`.


SanderMertens:
LGTM!

PR #1603, opened 2025-03-14
Add MSVC FORCEINLINE for always inline

Reddy-dev:


Reddy-dev:
Should be added now, added 1 basic test for testing if the id is less than component high id, not sure if you'd want more test if you can think of any.

SanderMertens:
Yeah I tried that in my first commit, but as you see it causes some builds to fail. I haven't investigated yet, but it looks like msvc doesn't export symbols for force_inline functions in some cases (which would be really bad).

SanderMertens:
Closing as CI is failing and no progress.

PR #1605, opened 2025-03-15
Added binary and hex base to script number

KenthJohan:
Supports binary and hex number in script.
Decimal: `123`
Hexadecimal `0xABC` (Added support)
Binary `0b01010101` (Added support)

This PR have some duplicate code.

Another approach would be to introduce more token enums like this:
* `EcsTokNumber` base 10
* `EcsTokNumberHex` base 16
* `EcsTokNumberBin` base 2


SanderMertens:
LGTM, thanks for the PR!

PR #1606, opened 2025-03-15
[meta] Opaque random-access serializer for vectors/arrays and structs

jpeletier:
This PR introduces two new opaque handlers that allow opaque implementers to provide random access to struct members or vector/array elements.

By defining `serialize_element` or `serialize_member`, the developer implementing the opaque can provide an interface to access an element or member directly.

For example, providing access to an `std::vector` item:

```cpp
// Reusable reflection support for std::vector
template<typename Elem, typename Vector = std::vector<Elem>>
flecs::opaque<Vector, Elem> std_vector_opaque(flecs::world &world) {
  return flecs::opaque<Vector, Elem>()
      .as_type(world.vector<Elem>())
      .count([](const Vector *data) { return data->size(); })
      .resize([](Vector *data, size_t size) { data->resize(size); })
      .ensure_element([](Vector *data, size_t elem) {
        if (data->size() <= elem) {
          data->resize(elem + 1);
        }
        return &data->data()[elem];
      })
     // new serialize_element handler:
      .serialize_element([](const flecs::serializer *s, const Vector *data,
                            size_t element) -> int {
        if (element >= data->size()) {
          return 1; // element not found
        }
        return s->value((*data)[element]); // invoke callback to submit the value.
      })
      .serialize([](const flecs::serializer *s, const Vector *data) {
        for (const auto &el : *data) {
          if (s->value(el)) {
            return 1;
          }
        }
        return 0;
      });
}
```

... or, to provide direct access to a struct member, use `serialize_member`:

```cpp
    world.component<Position>()
        .opaque(world.component()
            .member(flecs::F32, "x")
            .member(flecs::F32, "y"))
        .serialize_member([](
            const flecs::serializer * ser,
            const Position* data,
            const char* member) {
                const Position *p = (const Position*) data;
                if(!strcmp(member, "x")) {
                    return ser->value(p->x);
                } else if(!strcmp(member, "y")) {
                    return ser->value(p->y);
                }
                return 1;
            });
```

### DIFF
--- a/distr/flecs.h
+++ b/distr/flecs.h
@@ -19590,6 +19590,15 @@ using serialize_t = ecs_meta_serialize_t;
 template <typename T>
 using serialize = int(*)(const serializer *, const T*);
 
+/** Type safe variant of serialize_member function */
+template <typename T>
+using serialize_member = int(*)(const serializer *, const T*, const char* name);
+
+/** Type safe variant of serialize_element function */
+template <typename T>
+using serialize_element = int(*)(const serializer *, const T*, size_t element);
+
+
 /** Type safe interface for opaque types */
 template <typename T, typename ElemType = void>
 struct opaque {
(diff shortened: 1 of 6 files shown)


SanderMertens:
LGTM!

PR #1624, opened 2025-04-07
the memory of cond is not released correctly in windows

kagetu:
the memory of cond is not released correctly in windows


SanderMertens:
Oops. Thanks for the PR!

PR #1693, opened 2025-06-08
Fix typos in ComponentTraits.md and returning error code upon creating a struct with a syntax error but no members

999pingGG:
I noticed `ecs_meta_from_desc()` prints an error when passed the description `{` about missing the closing brace, but still returns 0 (success). This commit fixes that. Let me know about any issues with this! I'm not sure about how to add test cases for this tho.

### DIFF
--- a/distr/flecs.c
+++ b/distr/flecs.c
@@ -50393,7 +50393,8 @@ error:
 static
 const char * flecs_meta_utils_open_scope(
     const char *ptr,
-    flecs_meta_utils_parse_ctx_t *ctx)    
+    flecs_meta_utils_parse_ctx_t *ctx,
+    bool *error)
 {
     /* Skip initial whitespaces */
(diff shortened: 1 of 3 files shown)


SanderMertens:
Looks good! Can you add a test case (in `test/meta`) for the issue that is fixed by the 2nd commit?

999pingGG:
After sending this PR, I noticed some more issues related to empty structs even with my fix, and those require further investigation. I mean, you shouldn't be trying to parse and create empty structs, but resiliency is always good. I came across those bugs while writing my LuaJIT bindings and writing my own tests, let's prevent crashes as much as possible. I will investigate the problematic cases and write tests for them later.

SanderMertens:
Sounds good! Can you split off the changes to ComponentTraits.md into a separate PR?

SanderMertens:
Closing stale PR. Feel free to comment when there is progress and I'll reopen.

PR #1709, opened 2025-06-26
Update utils.hpp

AnonimiAngels:
This fix addresses a compatibility issue with GCC 4.9.2 compilers.

Specifically, older versions of GCC (like 4.9.2) do not fully support certain C++11 type traits, such as std::is_trivially_constructible, std::is_trivially_move_assignable, std::is_trivially_copy_assignable, std::is_trivially_copy_constructible, std::is_trivially_move_constructible, and std::is_trivially_copyable.

The provided fix conditionally defines these missing type traits within the std namespace when compiling with GCC versions older than 5.10. It implements them by leveraging the __is_pod(T) intrinsic, which checks if a type T is a Plain Old Data (POD) type. This approach ensures that code relying on these specific type traits can compile successfully on systems using GCC 4.9.2, maintaining broader compiler compatibility without requiring a full C++11 standard library implementation for these specific features. For newer GCC versions, the standard library's std:: implementations are used.


AnonimiAngels:
Tested with 



gcc --version
gcc (Debian 14.2.0-19) 14.2.0
Copyright (C) 2024 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.



And 



gcc (Debian 4.9.2-10+deb8u2) 4.9.2
Copyright (C) 2014 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

SanderMertens:
Hm technically I only support C++11 and above, but the fix seems easy enough. I won't be able to guarantee this keeps working in the future though, as I can't use older versions of gcc in CI.

The CI fail is because the amalgamated distr/flecs.h isn't updated (it has to match the source). You can either update it manually or have [bake](https://github.com/SanderMertens/bake) generate it for you. Once that's fixed it's good to go!

AnonimiAngels:
> Hm technically I only support C++11 and above, but the fix seems easy enough. I won't be able to guarantee this keeps working in the future though, as I can't use older versions of gcc in CI.
> 
> The CI fail is because the amalgamated distr/flecs.h isn't updated (it has to match the source). You can either update it manually or have [bake](https://github.com/SanderMertens/bake) generate it for you. Once that's fixed it's good to go!

Ill fix the amalgamated test

PR #1731, opened 2025-07-18
fix: bump cmake_minimum_required version

niceEli:
Previous versions to 3.10 is deprecated and cmake gives a warning

```
CMake Deprecation Warning at Coffee/vendor/flecs/CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 3.10 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
  to tell CMake that the project requires at least <min> but has been updated
  to work with policies introduced by <max> or earlier.
```


SanderMertens:
LGTM!

PR #1753, opened 2025-08-17
New API for supporting mut aliasing safety

Indra-db:
This PR introduces:

* locking for either component records (sparse components) or table-column locks (non-sparse components).
* separate per-thread / per-stage locking for table columns.
* everything gated by the `FLECS_MUT_ALIAS_LOCKS` define, so there is no performance impact when it’s not enabled.
* build tests and unit tests for the behaviour.

End users of this internal API must use it themselves and panic when the mut-alias guarantee is violated.

`ecs_lock_target_t` holds information about where the lock should be applied. For sparse components/IDs (sparse and/or non-fragmenting) it will use the `component_record` for locking. For table-column components it will lock the individual column.


Indra-db:
ci test, I expect a failure somewhere, I cannot run tests locally due to bake being problematic (see discord)

Indra-db:
all tests passing 🥳

Reddy-dev:
> ci test, I expect a failure somewhere, I cannot run tests locally due to bake being problematic (see discord)

@Indra-db could you link the message, I think i also have similar issues

Indra-db:
@Reddy-dev 

Tldr, Sander told me to use `bake run` as opposed to `bake test` which should fix the ouput

Indra-db:
@Reddy-dev this was my thread where I was dumping my experience (which I closed)
https://discord.com/channels/633826290415435777/1406452408149741770/1406998533449580585

Indra-db:
update to latest main

include/flecs/private/api_internals.h
```diff
@@ -333,6 +333,86 @@ FLECS_ALWAYS_INLINE ecs_table_t *flecs_table_traverse_add(
     ecs_id_t *id_ptr,
     ecs_table_diff_t *diff);
 
+#ifdef FLECS_SAFETY_LOCKS
+
+FLECS_API
+FLECS_ALWAYS_INLINE bool flecs_sparse_id_record_lock_read_begin(
```
SanderMertens, line 339: What is a "sparse id"? Can you explain this in the PR description?

SanderMertens, line 339: Since all these functions accept a component record, shouldn't this be in storage/component_index.h?

SanderMertens: Ah it's a sparse "id record". This is now called a component record. Can you rename this to:

```
flecs_component_record_read_begin
```

include/flecs/private/api_internals.h
```diff
@@ -333,6 +333,86 @@ FLECS_ALWAYS_INLINE ecs_table_t *flecs_table_traverse_add(
...
+FLECS_API
+FLECS_ALWAYS_INLINE bool flecs_sparse_id_record_lock_write_begin(
+    ecs_component_record_t *idr);
+
+FLECS_API
+FLECS_ALWAYS_INLINE bool flecs_sparse_id_record_lock_write_end(
+    ecs_component_record_t *idr);
+
+FLECS_API
+FLECS_ALWAYS_INLINE bool flecs_table_column_lock_read_begin(
```
SanderMertens: Can you rename this to `flecs_table_column_read_begin`

include/flecs/private/api_internals.h
```diff
@@ -333,6 +333,86 @@ FLECS_ALWAYS_INLINE ecs_table_t *flecs_table_traverse_add(
...
+    ecs_table_t *table,
+    const int16_t column_index);
+
+FLECS_API
+FLECS_ALWAYS_INLINE bool flecs_table_column_lock_write_end(
+    ecs_table_t *table,
+    const int16_t column_index);
+
+FLECS_API
+FLECS_ALWAYS_INLINE bool flecs_sparse_id_record_lock_read_begin_multithreaded(
```
SanderMertens: Can this be merged with the previous function and an additional argument?

SanderMertens:
I'm still seeing performance regressions when comparing with latest master:

Screenshot 2025-12-06 at 2 44 11 PM

I think it's better to separate out the implementation into its own functions so there's no risk of regressions.

(Note: benchmarks were done without the `FLECS_MUT_ALIAS_LOCKS` flag on a regular Flecs master)

src/storage/component_index.c
```diff
@@ -1246,3 +1246,83 @@ void flecs_component_update_childof_depth(
 
     flecs_component_update_childof_w_depth(world, cr, new_depth);
 }
+
+#ifdef FLECS_MUT_ALIAS_LOCKS
+
+int32_t flecs_sparse_id_record_lock_inc(
```
SanderMertens: This is a bunch of functions that do basically just wrap each other and do nothing. Ugly code.

src/storage/table.c
```diff
@@ -2970,3 +2989,233 @@ char* ecs_table_str(
         return NULL;
     }
 }
+
+#ifdef FLECS_MUT_ALIAS_LOCKS
+
+void flecs_table_init_column_locks(
```
SanderMertens: Can we put this code in a separate table_column_locks file? It's a lot of code that won't be enabled for most applications. Putting an #ifdef FLECS_MUT_ALIAS_LOCKS in the middle of the table.c file makes the code hard to read.

src/storage/table.c
```diff
@@ -2970,3 +2989,233 @@ char* ecs_table_str(
...
+
+            if (table->column_lock) {
+                flecs_wfree_n(world, int32_t, previous_stage_count * table->column_count, table->column_lock);
+                table->column_lock = NULL;
+            }
+        }
+    }
+}
+
+int32_t flecs_table_column_lock_inc(
```
SanderMertens: Same function churn. Ugly, needs a better API.

src/entity.c
```diff
@@ -135,6 +136,82 @@ void* flecs_get_base_component(
     return NULL;
 }
 
+static
```
SanderMertens: This code looks like a duplicated version of the existing function.

SanderMertens (changes requested):

include/flecs/private/api_internals.h
Indra-db: according to what I read with rust FFI with C, the branch wouldn't be able to get eliminated, so the answer is no. Locks already degrade the performance, so I would want to prevent it from degrading even more

PR #1763, opened 2025-08-21
Assert for char ptr

Reddy-dev:
I forget what issue this caught but this did catch an issue at some point in my testing


Reddy-dev:
why would the amalgamated fail on this? they are the same?

SanderMertens:
Maybe a whitespace issue? That check only fails if there's a difference.

Reddy-dev:
@SanderMertens Should be fixed now

PR #1768, opened 2025-08-22
Update docs/EntitiesComponents.md to change `is_enabled` to `enabled`

ericjeker:
In C++, the function for checking if a component is enabled is called `enabled` and not `is_enabled`.


SanderMertens:
LGTM!

PR #1779, opened 2025-09-01
get_constant with an entity input param

Reddy-dev:


SanderMertens:
What is the purpose of this function? Since you're passing in the type as well, can't you just use that when getting the target?

It seems like this is used to obtain a constant of a different enum type, but that doesn't seem safe.

Reddy-dev:
@SanderMertens > What is the purpose of this function? Since you're passing in the type as well, can't you just use that when getting the target?
> 
> It seems like this is used to obtain a constant of a different enum type, but that doesn't seem safe.

``constant of a different enum type`` this is meant to get the runtime constant that inherits from an integer type, as in just a wrapper over the runtime variants of the equivalent functiuons in the non-runtime get_constant

PR #1788, opened 2025-09-08
Small compile time improvements

dementive:
- Adds a `conditional` template that is faster to instantiate than std::conditional. I got this from here: https://github.com/facebook/fatal/blob/main/fatal/type/conditional.h and adapted it to my code and have been using it for awhile now here: https://github.com/dementive/lt/blob/main/src/lt/types/conditional.hpp.


SanderMertens:
LGTM!

PR #1808, opened 2025-09-25
[bugfix] flecs_rtt_vector_equals

StrailerPark:


SanderMertens:
Can you add a test case that reproduces the issue? See the quickstart for how to run tests:
https://www.flecs.dev/flecs/md_docs_2Quickstart.html#running-tests-bake

Running bake will also regenerate the amalgamated files which fixes the failing CI job.

PR #1813, opened 2025-09-28
fix(core): Correct DbgHelp.h capitalisation for MinGW cross-compilation

### LINKED ISSUE #1812: Cross-compilation for Windows fails on Linux due to DbgHelp.h capitalisation
**Describe the bug**

Hello!

I've been trying to cross-compile Flecs for Windows from a Linux host, and I've run into a build issue that I've spent a fair bit of time digging into. The build fails with a "file not found" error for the `<DbgHelp.h>` header.

After a lot of investigation, I discovered the root cause is a case-sensitivity mismatch. The Flecs source code correctly uses the canonical Windows capitalisation (`DbgHelp.h`), but it turns out that the standard MinGW packages on all major Linux distributions (I checked Fedora, Debian, and Arch) provide this file and its corresponding library in all-lowercase (`dbghelp.h` and `libdbghelp.a`).

Since Linux filesystems are case-sensitive, the compiler can't find the requested header file. After fixing that, a subsequent linker error occurs for the same reason. The Flecs `CMakeLists.txt` makes this dependency mandatory for all Windows builds, so it's currently not possible to cross-compile Flecs out of the box in a standard Linux MinGW environment.

**To Reproduce**

Here are the steps I used to create a minimal test case that reproduces the behavior:

1.  **Environment:** Set up a Linux system (I used Fedora 42) with the MinGW-w64 toolchain (`mingw64-gcc`, `mingw64-headers`).
2.  **Clone:** `git clone --recurse-submodules https://github.com/SanderMertens/flecs.git`
3.  **Toolchain File:** Create a `toolchain-windows.cmake` for cross-compilation:
    ```cmake
    set(CMAKE_SYSTEM_NAME Windows)
    set(CMAKE_C_COMPILER x86_64-w64-mingw32-gcc)
    ```
4.  **Configure:** Run CMake to set up the build.
    ```bash
    cmake -S . -B build -DCMAKE_TOOLCHAIN_FILE=toolchain-windows.cmake
    ```
5.  **Build:** Attempt to compile the project.
    ```bash
    cmake --build build
    ```
6.  **See Error:** The build fails with the following error:
    ```
    fatal error: DbgHelp.h: No such file or directory
       |          ^~~~~~~~~~~
    compilation terminated.
    ```

**Expected behavior**

I expected the project to compile and link successfully without any errors.

**Additional context**

Just to add some extra context from my research, this seems to be the de facto standard for how MinGW headers and libraries are packaged on Linux, so it's less of a bug in any one distribution and more of a general incompatibility with the cross-compilation environment.

A complete solution requires fixing this case mismatch in **two places**:

1.  **In the source code (`src/os_api.c`):** The `#include` needs to be changed to lowercase: `<dbghelp.h>`.
2.  **In the build script (`CMakeLists.txt`):** The `target_link_libraries` command needs to link against the lowercase `dbghelp` library instead of `dbgHelp`.

I hope this report is helpful and provides a full picture of the issue!

*   **OS:** Fedora 42 (in a Toolbx container)
*   **Compiler:** `x86_64-w64-mingw32-gcc`
*   **Toolchain:** MinGW-w64 (from Fedora repositories)

thomashodgkinson1987:
This change modifies the include to use lowercase, which matches the de facto standard for MinGW header packaging on all major Linux distributions (Fedora, Debian, Arch). This resolves a case-sensitivity build failure when cross-compiling.

Closes #1812


SanderMertens:
LGTM! Can you also update distr/flecs.c with the change? That will fix the CI failure.

SanderMertens:
Thanks!

PR #1834, opened 2025-10-11
Add blacklist callback to serializer descriptor

Real-Septicake:
As the title says, this pr adds a callback to the serializer descriptors to determine if a component should be serialized.

A few tests have also been added to make sure the feature functions properly, checking a callback that:
- Allows all components
- Allows no components
- Allows some components

Tests were run locally in a Linux environment


SanderMertens:
Looks good! I think renaming the field to `.component_filter` or sth is better, but otherwise 👍

Real-Septicake:
Yeah, I was pretty sure there'd be something better for that. Will do

SanderMertens:
LGTM!

PR #1891, opened 2025-12-12
added get_version to get generation from entity

tabdroid-1:
```cpp
    flecs::world world {};
    flecs::entity entity = m_World.entity();
    std::cout << "Entity id: " << entity << " Entity version: " << world.get_version(entity) << "\n";
    entity.destruct();

    entity = m_World.entity();
    std::cout << "Entity id: " << entity << " Entity version: " << world.get_version(entity) << "\n";
    entity.destruct();

    entity = m_World.entity();
    std::cout << "Entity id: " << entity << " Entity version: " << world.get_version(entity) << "\n";
    entity.destruct();

    entity = m_World.entity();
    std::cout << "Entity id: " << entity << " Entity version: " << world.get_version(entity) << "\n";
    entity.destruct();
```
output

```
Entity id: 578 Entity version: 0
Entity id: 4294967874 Entity version: 1
Entity id: 8589935170 Entity version: 2
Entity id: 12884902466 Entity version: 3
```

### DIFF
--- a/distr/flecs.c
+++ b/distr/flecs.c
@@ -10445,6 +10445,7 @@ void ecs_set_version(
     }
 }
 
+
 uint32_t ecs_get_version(
     ecs_entity_t entity)
 {
--- a/distr/flecs.h
+++ b/distr/flecs.h
@@ -23608,7 +23608,7 @@ struct world {
         ecs_set_version(world_, e);
     }
 
-    /** Get version of entity with generation provided.
+    /** Get version of provided entity.
      * 
      * @see ecs_get_version()
      */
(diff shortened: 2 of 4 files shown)


zachrammell:
What's the rationale behind having `get_version` be a method of `world` in the C++ API if the equivalent procedure in C doesn't need a world? It would probably be more useful as a method of `entity`, no?

tabdroid-1:
> What's the rationale behind having `get_version` be a method of `world` in the C++ API if the equivalent procedure in C doesn't need a world? It would probably be more useful as a method of `entity`, no?

My rationale was if world has set_version then user will probably think there is get_version. Which was the case for me

SanderMertens:
LGTM!

PR #1898, opened 2025-12-26
docs/EntitiesComponents.md: minor fixes

ludolpif:
Hi,
I am a french enthusiast mocking up a streaming software out of SDL3, flecs and ImGui for no good reasons.

I think I have spotted minor details that can be improved in docs/EntitiesComponents.md and here a PR.
- a missing `;`
- quotes seems to cancel back-ticks in the resulting https://www.flecs.dev/flecs/md_docs_2EntitiesComponents.html
- `set` mentioned on `assign` line (previous line duplicate ?), hoping it correct to say the same as for `set` 

Your work seems very impressive!
Thanks for all the fish,
Ludolpif

### DIFF
--- a/docs/EntitiesComponents.md
+++ b/docs/EntitiesComponents.md
@@ -1174,7 +1174,7 @@ e.add::<flecs::Disabled>();
 </div>
 
 ## Components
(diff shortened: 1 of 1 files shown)


ludolpif:
Just added an additionnal comment fix, it seems the original was copy-pasted from the previous section, leaved unchanged. Taken the one from the C++ example.

SanderMertens:
Looks good! Removing the quotes does make it a bit harder to read though, I would rather remove the backticks.

ludolpif:
Hi,
Sorry for delay, I removed backticks instead of quotes as last comment suggests.
Have a nice day,

SanderMertens:
LGTM!

PR #1899, opened 2026-01-01
Repro/crlf script loading

60k41p:
Repro only; not meant to be merged.

# How to reproduce

## Input FlecsScript

I have included test.{CRLF|LF}.flecs that are encoded using the different line endings, the contents of which is just the first example snippet on https://www.flecs.dev/flecs/md_docs_2FlecsScript.html.

Note that `git` may automatically convert the line endings during file check-in. We don't want that to happen during this repro, so it's probably a good idea to run `git config core.autocrlf false` to turn that off.

If needed, on Linux & macOS use `unix2dos -v` to re-generate the CRLF version of the file. This tool can be installed via `brew install dos2unix` on macOS.

## Reproducing

```
$ ./compile.sh
+ gcc-15 -c ../../distr/flecs.c -std=c99 -O2 -o flecs.o
+ g++-15 repro.cpp flecs.o -o repro -std=c++17 -lpthread

a@n8:~/src/flecs/_REPRO/crlf-script-loading$ ./repro test.CRLF.flecs
error: test.CRLF.flecs: 4: expected '
'
  value = f32
            ^
Error running script

a@n8:~/src/flecs/_REPRO/crlf-script-loading$ ./repro test.LF.flecs
Script 'test.LF.flecs' executed successfully.
```

### DIFF
--- a/_REPRO/crlf-script-loading/build-repro.sh
+++ b/_REPRO/crlf-script-loading/build-repro.sh
@@ -1,0 +1,5 @@
+#!/usr/bin/env bash
+set -ex
+
+gcc-15 -c ../../distr/flecs.c -std=c99 -O2 -o flecs.o
+g++-15 repro.cpp flecs.o -o repro -std=c++17 -lpthread
--- a/_REPRO/crlf-script-loading/repro.cpp
+++ b/_REPRO/crlf-script-loading/repro.cpp
@@ -1,0 +1,21 @@
+#include <iostream>
+#include "../../distr/flecs.h"
+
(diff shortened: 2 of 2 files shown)


60k41p:
Sorry, I meant to open this PR on my own fork.

I may open a fix PR here separately.

60k41p:
See https://github.com/VoxInteractive/flecs/pull/1

60k41p:
Fix: https://github.com/SanderMertens/flecs/pull/1900
Seems to have had a false failure on one of the CI checks.

PR #1900, opened 2026-01-01
Fix: Handle CRLF line endings in Flecs scripts

60k41p:
This resolves an issue where Flecs scripts would fail to parse if they used CRLF (Carriage Return Line Feed) line endings. The script parser was expecting only LF (Line Feed) as a line ending, leading to parsing errors when an unexpected carriage return character was encountered.

The fix involves modifying the  function in
 to normalize line endings. After loading the file content,
all carriage return characters () are removed, ensuring that the script parser only receives LF line endings.

A new helper function flecs_str_remove_all_occurrences_of_char was added to facilitate this normalization.

This change ensures that Flecs scripts can be loaded and executed correctly regardless of their line ending convention.

For repro of the bug, please see https://github.com/VoxInteractive/flecs/pull/1

After this change:
```
a@n8:~/src/flecs/_REPRO/crlf-script-loading$ ./repro test.LF.flecs
Script 'test.LF.flecs' executed successfully.
a@n8:~/src/flecs/_REPRO/crlf-script-loading$ ./repro test.CRLF.flecs
Script 'test.CRLF.flecs' executed successfully.
```

### DIFF
--- a/test/script/src/Eval.c
+++ b/test/script/src/Eval.c
@@ -133,7 +133,7 @@ void Eval_multiple_trailing_crlfs(void) {
 void Eval_trailing_crlfs_around_struct(void) {
     ecs_world_t* world = ecs_init();
 
-    test_assert(ecs_script_run(world, NULL, "using flecs.meta\r\n \r\nstruct MaxSpeed {\r\n  value = f32\r\n}", NULL) == 0);
+    test_assert(ecs_script_run(world, NULL, "using flecs.meta\r\n\r\nstruct MaxSpeed {\r\n  value = f32\r\n}\r\n", NULL) == 0);
 
     ecs_fini(world);
 }


SanderMertens:
Thanks for the PR!

What this needs in order to get merged is:
- A test case that demonstrates the issue. You can prevent git from removing `\r` by parsing a flecs script string, and using the `\r\n` escape sequence.
- The fix removes all `\r`'s from a script which works, but is not very efficient. A better solution would be to add `\r` support to the tokenizer directly.
- The CI failure is because the amalgamated file is out of sync - it needs to match the source.

SanderMertens:
There are a number of tests that test for `\r\n` but they don't seem to be failing, which is odd: https://github.com/SanderMertens/flecs/blob/master/test/script/src/Eval.c#L80

60k41p:
> Thanks for the PR!
> 
> What this needs in order to get merged is:
> 
>     * A test case that demonstrates the issue. You can prevent git from removing `\r` by parsing a flecs script string, and using the `\r\n` escape sequence.
> 
>     * The fix removes all `\r`'s from a script which works, but is not very efficient. A better solution would be to add `\r` support to the tokenizer directly.
> 
>     * The CI failure is because the amalgamated file is out of sync - it needs to match the source.

Hi,
I haven't been able to reproduce this in [tests](https://github.com/VerdantInteractive/flecs/blob/33db504243b2be77897f85e85589f2b2df9e659f/test/script/src/Eval.c#L133). Which is a bit perplexing.

Thank you for the info - I'll keep that in mind if/when I open my next PR.

Issue #1703, opened 2025-06-19
Retrieving multiple (mutable) components for an entity in 1 go without a lambda


BasBuur:
**Describe the problem you are trying to solve.**
Entt allows retrieving n components in 1 go with the following syntax:

```cpp
auto & [pos, vel] = registry.get<Position, Velocity>(entity);
```

This is a really nice feature and it would be great if it could be added to flecs. I know flecs allows retrieving multiple components with a lambda but I prefer the components in my local scope without a lambda.

Bonus points if we could get a mutable version as well. For example `entity.get_all<Position, Velocity>()` and `entity.get_all_mut<Position, Velocity>()`.

SanderMertens:
I'm planning to migrate the API to C++17 which will enable this API. I'm expecting to begin working on this within a month.

SimonPiCarter:
I am toying a bit with some implementation to solve this but from my understanding allowing the usage of your syntax would require flecs to define some std template specialization.

Would something along the line of 
```c++
auto foo = e.get_all<Position, Velocity>();
Position const &pos = foo.get<0>();
```
Be acceptable? 
I don't really see how this would be better than just fetching the components one by one tho.

An alternative would be to have the following options: get_all_2, get_all_3 with fixed numbers up to an arbitrary limit but to me that would be cumbersome and not worth extending the api. There may be some trickery to make them all have the same name but there would still be an arbitrary limit on how many you could fetch at once.

SimonPiCarter:
@SanderMertens would an api providing this functionality but with an arbitrary limit on how many components you can get at once be fine?
It would add quite a bit of boilerplate code to get it working but we could have exactly the syntax suggested.
Compilation errors would arise if a user called get_all with more than M components.

SanderMertens:
@SimonPiCarter what is the reason for the limitation? I believe EnTT supports structured bindings, I'd take a look at how it's supported there.

Also the aforementioned switch to C++17 has happened now, so there's no need to keep it compatible with C++11 anymore.

SimonPiCarter:
To support limitless structure binding it is required to define some std templates (cf https://en.cppreference.com/w/cpp/language/structured_binding.html see Case 2)

Iirc std usage is non desired in flecs therefore it is not possible to do like enTT (which use the STL heavily)

EDIT If you're fine  with using std::tuple then we can have this pretty easily
EDIT2 without using std::tuple the implementation is a bit more tedious but an incude to \<utility\> should be enough

[...]

PR #1930, opened 2026-02-05
FIX: Added component tuple getters in cpp api

SimonPiCarter:
Addressing https://github.com/SanderMertens/flecs/issues/1703

This a proposal with no added include and a limitation on max component fetched

### DIFF
--- a/distr/flecs.h
+++ b/distr/flecs.h
@@ -26199,7 +26199,7 @@ struct entity_view : public id {
     }
 
     template<typename... Ts>
-    auto try_gefdsfdst_mut_n() const {
+    auto try_get_mut_n() const {
         flecs_static_assert(sizeof...(Ts) > 1, "try_get_mut_n requires at least two components");
         flecs_static_assert(sizeof...(Ts) < 9, "try_get_mut_n cannot fetch more than eight components");
         return typename tuple_builder<sizeof...(Ts), Ts...>::type_ptr {try_get_mut<Ts>()...};


SimonPiCarter:
Tests and example have been added, some comments and documentation is still required

SanderMertens:
That looks pretty good! The only nit I have is with the naming, could we maybe change `get_all` into `get_n`? `get_all` sounds more like you're getting all components from the entity.

SimonPiCarter:
Surely I'll change that, you are right.

SimonPiCarter:
What do you think about also naming it get without any suffix, that would also be a possibility

SanderMertens:
I don't think you can do that because of pairs:
```
const Eats& e = e.get<Eats, Fruit>();
```

SimonPiCarter:
Ah yes that's true. get_n it is then

SimonPiCarter:
Everything should be good now. I also fixed a missed conf file modification.

FYI maybe its missing a check to detect modification of sources after builds for test conf (thinking test/sc/srcmain.cpp here), since I have overridden  the file without seeing the top comment which ended with my change to be overridden during build.

You have such check for the main sources but not for the tests if I understand correctly.

A full test build of the tests in `test-amalgamated` should be enough?

EDIT last thing is adding some comments to the methods, will do later

SanderMertens:
LGTM! Follow up steps for this could be to implement a more optimized version of this, since you can avoid lookups if you're fetching components from the same entity.

> You have such check for the main sources but not for the tests if I understand correctly.

I could look into that yeah. It doesn't come up as often because the typical workflow is:

- add test case to JSON
- call `bake run test/<suite>`
- files get updated

but I agree that if you're not using bake (which is probably most people) this would be nice to have.

PR #1958, opened 2026-02-21
Call init_builtin_components() at the end of world reset

60k41p:
Fixes `fatal: flecs.h: 30083: assert(registered(world)): component 'EcsRest' must be registered before use (INVALID_OPERATION)` on `world.set<flecs::Rest>({});` following a `world.reset()`.


SanderMertens:
It looks like this PR got mixed up with pipeline disabling. Can you remove the pipeline changes from this PR, and include a test case that repros the issue?

SanderMertens:
LGTM!

PR #1959, opened 2026-02-21
Implement Disabling of Pipelines

60k41p:
# Changes
- Early return in ecs_run_pipeline when a pipeline is disabled.
- Simplifiy ecs_progress() by delegating thread management and pipeline execution to ecs_run_pipeline().
- Add tests to verify that the systems in disabled default or custom pipelines don't execute

# Reproducer
```
#include <iostream>
#include <flecs.h>

int main() {
    // Create a Flecs world
    flecs::world world;

    // Create a simple system that prints whenever it runs
    world.system<>()
        .run([](flecs::iter &it) {
            std::cout << "[system] executed" << std::endl;
        });

    // Get pipeline entity and print its initial state
    flecs::entity pipeline = world.get_pipeline();
    std::cout << "pipeline id: " << pipeline.id() << ", valid: " << pipeline.is_valid()
              << ", enabled: " << pipeline.enabled() << std::endl;

    // First progress call: system should run
    std::cout << "--- First progress() ---" << std::endl;
    world.progress();

    // Disable the pipeline entity
    std::cout << "Disabling pipeline (pipeline.disable())" << std::endl;
    pipeline.disable();

    std::cout << "pipeline enabled after disable(): " << pipeline.enabled() << std::endl;

    // Subsequent progress calls: check whether system still runs
    std::cout << "--- Second progress() ---" << std::endl;
    world.progress();

    std::cout << "--- Third progress() ---" << std::endl;
    world.progress();

    std::cout << "Done." << std::endl;
    return 0;
}
```

## Before

```
pipeline id: 400, valid: 1, enabled: 1
--- First progress() ---
[system] executed
Disabling pipeline (pipeline.disable())
pipeline enabled after disable(): 0
--- Second progress() ---
[system] executed
--- Third progress() ---
[system] executed
Done.
```

## After
```
pipeline id: 400, valid: 1, enabled: 1
--- First progress() ---
[system] executed
Disabling pipeline (pipeline.disable())
pipeline enabled after disable(): 0
--- Second progress() ---
--- Third progress() ---
Done.
```

### DIFF
--- a/distr/flecs.c
+++ b/distr/flecs.c
@@ -62190,6 +62190,10 @@ void ecs_run_pipeline(
         pipeline = world->pipeline;
     }
 
+    if (ecs_has_id(world, pipeline, EcsDisabled)) {
+        return;
+    }
+
     /* create any worker task threads request */
     if (ecs_using_task_threads(world)) {
         flecs_create_worker_threads(world);
@@ -62197,8 +62201,11 @@ void ecs_run_pipeline(
 
     EcsPipeline *p = 
         ECS_CONST_CAST(EcsPipeline*, ecs_get(world, pipeline, EcsPipeline));
+    ecs_check(p != NULL, ECS_INVALID_OPERATION,
+        "pipeline entity is missing flecs.pipeline.Pipeline component");
     flecs_workers_progress(world, p->state, delta_time);
 
+error:
     if (ecs_using_task_threads(world)) {
         /* task threads were temporary and may now be joined */
         flecs_join_worker_threads(world);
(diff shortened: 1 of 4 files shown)


SanderMertens:
I'm closing this PR, because (copied from Discord):

> Pipeline disabling is not a thing, because you explicitly call either progress() or run_pipeline(). The reasoning is that if you don't want to run the pipeline, you shouldn't call those functions (because they don't really have any other purpose)

> What you can do however is disable the phase or module that systems/observers are defined in. That does work

PR #2035, opened 2026-03-25
query: use actual table id for IsA matches in fixed traversal

60k41p:
Updates `flecs_query_trav_fixed_src_up_fixed_second` to use the actual component ID stored in the table's type array when a match is found for an `EcsIsA` relationship, rather than reconstructing the pair from arguments.

[Bug repro](https://discord.com/channels/633826290415435777/1455553733978099763/1456850832355102720)

Changes:
- Capture the column index returned by `ecs_search_relation`.
- Add a check for `index != -1` to ensure the relationship was found in the table (Self) before accessing the type array.
- When `trav == EcsIsA`, set the matched ID to `table->type.array[column]`.
- Manually populate the iterator fields and variables with the resolved `matched` ID.

This ensures that the query iterator reports the exact component ID present in the archetype when matching `IsA` relationships.

Also, this PR updates the link to the Stagehand repository (which got renamed from VoxInteractive/Godot-Flecs).

### DIFF
--- a/distr/flecs.c
+++ b/distr/flecs.c
@@ -85722,10 +85722,10 @@ bool flecs_query_trav_fixed_src_up_fixed_second(
 
     /* Check if table has transitive relationship by traversing upwards */
     ecs_table_record_t *tr = NULL;
-    ecs_search_relation(ctx->world, table, 0, 
-        ecs_pair(trav, second), trav, EcsSelf|EcsUp, NULL, NULL, &tr);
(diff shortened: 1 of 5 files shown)


SanderMertens:
Can you add a test case that reproduces the failure that this fixes?

60k41p:
> Can you add a test case that reproduces the failure that this fixes?

Test added.

PR #2052, opened 2026-04-10
query_builder::each const correctness

Cobryis:


Cobryis:
@SanderMertens can you retrigger the two failed tests? I don't think those had anything to do with my change.

PR #2072, opened 2026-04-30
Fix ecs_import MODULE_UNDEFINED assert when loading modules dynamically

### LINKED ISSUE #1755: Assert error when trying to load a module at runtime
**Describe the bug**
Dynamic loading of modules doesn't work.

**To Reproduce**
`main.c` file:
```c
#include <stdio.h>
#include <dlfcn.h>

#include "flecs.h"

ecs_os_dl_t my_api_dlopen(const char *libname) {
    void *handle = dlopen(libname, RTLD_NOW|RTLD_LOCAL);
    if (!handle) {
        fprintf(stderr, "%s\n", dlerror());
        exit(EXIT_FAILURE);
    }

    return (ecs_os_dl_t)handle;
}

ecs_os_proc_t my_api_dlproc(ecs_os_dl_t lib, const char *procname) {
    return (ecs_os_proc_t)dlsym((void*)lib, procname);
}

void my_api_dlclose(ecs_os_dl_t lib) {
    dlclose((void*)lib);
}

int main(void) {
    ecs_os_set_api_defaults();
    ecs_os_api.dlopen_ = my_api_dlopen;
    ecs_os_api.dlproc_ = my_api_dlproc;
    ecs_os_api.dlclose_ = my_api_dlclose;
    ecs_os_set_api(&ecs_os_api);

    ecs_world_t *ecs = ecs_init();
    ecs_import_from_library(ecs, "window", NULL);

    ecs_app_run(ecs, &(ecs_app_desc_t){});
    ecs_fini(ecs);
}
```

`window.c` file:
```c
#include "flecs.h"

void WindowImport(ecs_world_t *world) {
  // fatal: flecs.c: 26927: assert: e != 0 Window (MODULE_UNDEFINED)
  ECS_MODULE(world, Window);

  // For some reason, registering a tag or component with the module name fixes the error
  // ECS_TAG(world, Window);
}
```

Get an fatal error:
```terminal
yahton@debian:~/atomix$ ./nob && ./build/main 
[INFO] directory `./build` already exists
[INFO] directory `./build/libs` already exists
[INFO] directory `./build/modules` already exists
[INFO] CMD: cc -Wall -Wextra -shared -fPIC -o ./build/libs/librgfw.so -x c -DRGFW_IMPLEMENTATION -DRGFW_OPENGL ./include/RGFW.h -lX11 -lXrandr -lGL
[INFO] CMD: cc -Wall -Wextra -I./include -L./build/libs -shared -fPIC -o ./build/modules/libwindow.so ./modules/window.c -lflecs -lrgfw
[INFO] CMD: cc -Wall -Wextra -I./include -L./build/libs -o ./build/main main.c -lflecs
fatal: flecs.c: 27215: assert(e != 0): Window (MODULE_UNDEFINED)
./build/libs/libflecs.so(+0x57931) [0x7fcbceff1931]
./build/libs/libflecs.so(ecs_printv_+0xe5) [0x7fcbcf00f06e]
./build/libs/libflecs.so(ecs_log_+0xc4) [0x7fcbcf00f277]
./build/libs/libflecs.so(ecs_assert_log_+0x101) [0x7fcbcf00fae4]
./build/libs/libflecs.so(ecs_import+0x2a5) [0x7fcbcf0158ca]
./build/libs/libflecs.so(ecs_import_from_library+0x57a) [0x7fcbcf015f0e]
./build/main(+0x12f1) [0x5647cf6c82f1]
/lib/x86_64-linux-gnu/libc.so.6(+0x2724a) [0x7fcbcedc524a]
/lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0x85) [0x7fcbcedc5305]
./build/main(+0x1121) [0x5647cf6c8121]
Aborted
```

**Expected behavior**
No assert fatal error :/

**Additional context**
Debian GNU/Linux 12 (bookworm) x86_64

**My thoughts**
I am new to ECS, but it seems that `ECS_MODULE` does not register module entity, even though `ecs_import` expects it.

### LINKED ISSUE #2071: `ecs_import_from_library` fails with MODULE_UNDEFINED assert after successful module initialization
### Summary

`ecs_import_from_library()` successfully loads the dylib and calls the module's import function, but then fails with a `MODULE_UNDEFINED` assert because the internal lookup name doesn't match the name registered by `ECS_MODULE`.

### Environment

- Flecs v4.1.5 (amalgamation)
- macOS 15, Apple Silicon (aarch64)
- C++17, Apple Clang

### Steps to Reproduce

**Plugin (physics_plugin.cpp → libphysics_plugin.dylib):**

```cpp
#include "flecs.h"

extern "C" void PhysicsPluginImport(ecs_world_t *world) {
    ECS_MODULE(world, PhysicsPlugin);
    // ... register components/systems ...
}
```

**Host:**

```cpp
#include "flecs.h"
#include <dlfcn.h>

int main() {
    // Set up OS API dl callbacks
    ecs_os_api.dlopen_ = [](const char* n) -> ecs_os_dl_t { return (ecs_os_dl_t)dlopen(n, RTLD_NOW); };
    ecs_os_api.dlproc_ = [](ecs_os_dl_t l, const char* n) -> ecs_os_proc_t { return (ecs_os_proc_t)dlsym((void*)l, n); };
    ecs_os_api.dlclose_ = [](ecs_os_dl_t l) { dlclose((void*)l); };

    // This calls PhysicsPluginImport successfully, then asserts
    ecs_import_from_library(world, "physics.plugin", NULL);
    
    ecs_fini(world);
    return 0;
}
```

**Build:**

```bash
# Plugin
c++ -std=c++17 -O2 -shared -undefined dynamic_lookup physics_plugin.cpp -o libphysics_plugin.dylib

# Host (statically links flecs)
c++ -std=c++17 -O2 main.cpp flecs.o -o test
DYLD_LIBRARY_PATH=. ./test
```

### Expected Behavior

Module loads successfully from dylib.

### Actual Behavior

The import function `PhysicsPluginImport` is called and runs to completion (verified with printf). Then `ecs_import_from_library` asserts:

```
fatal: flecs.c: 27780: assert(e != 0): PhysicsPlugin (MODULE_UNDEFINED)
```

### Root Cause Analysis

`ecs_import_from_library("physics.plugin", NULL)` does the following internally:

1. Derives import function name: `"physics.plugin"` → `"PhysicsPluginImport"` ✅
2. Derives library filename: `"physics.plugin"` → `"libphysics_plugin.dylib"` ✅  
3. Loads library and calls `PhysicsPluginImport(world)` ✅
4. Derives module lookup name: `"physics.plugin"` → `"PhysicsPlugin"` 
5. Calls `ecs_lookup_symbol(world, "PhysicsPlugin", ...)` → **returns 0** ❌

The problem is in step 5. `ECS_MODULE(world, PhysicsPlugin)` internally calls `flecs_module_path_from_c("PhysicsPlugin")` which converts it to `"physics.plugin"` (dot-separated path). The entity is registered with path `"physics.plugin"` (parent entity `"physics"`, child entity `"plugin"`).

But `ecs_import_from_library` looks up `"PhysicsPlugin"` (PascalCase) — which doesn't exist as either an entity name or a symbol.

**Verification:**

```cpp
ecs_entity_t m = ecs_import_c(world, PhysicsPluginImport, "PhysicsPlugin");
// Returns 496 ✅ (ecs_import_c works because it returns scope directly, no lookup needed)

ecs_lookup(world, "PhysicsPlugin");       // → 0 ❌
ecs_lookup(world, "physics.plugin");      // → 496 ✅
ecs_lookup_symbol(world, "PhysicsPlugin", true, false);  // → 0 ❌
```

### Workaround

Use manual `dlopen`/`dlsym` instead of `ecs_import_from_library`:

```cpp
void* handle = dlopen("libphysics_plugin.dylib", RTLD_NOW);
auto init_fn = (void(*)(ecs_world_t*))dlsym(handle, "PhysicsPluginImport");
ecs_import_c(world, init_fn, "PhysicsPlugin");  // works!
```

### Suggested Fix

In `ecs_import_from_library` (flecs.c ~line 27780), after calling the import function, the module should be looked up using the dot-separated path (same conversion that `ECS_MODULE` uses) rather than the raw PascalCase module name. Something like:

```c
// Current (broken):
ecs_entity_t e = ecs_lookup_symbol(world, module, true, false);

// Should be:
char *module_path = flecs_module_path_from_c(module);
ecs_entity_t e = ecs_lookup_symbol(world, module_path, true, false);
ecs_os_free(module_path);
```

Or alternatively, `ecs_import_from_library` could use the same approach as `ecs_import_c` — capture the scope entity directly from the import function execution rather than doing a post-hoc lookup.

bg5sbk:
The second lookup in ecs_import() used the raw module_name (PascalCase) instead of the converted path (dot.separated). This caused ecs_import_from_library() to always fail with MODULE_UNDEFINED because ECS_MODULE registers the entity with the dot-separated path (e.g. 'physics.plugin'), but the post-import lookup searched for the PascalCase name (e.g. 'PhysicsPlugin') which doesn't exist.

Fix: use the already-converted 'path' variable for both lookups, and defer ecs_os_free(path) to the end of the function.

Fixes #2071
Fixes #1755

### DIFF
--- a/src/addons/module.c
+++ b/src/addons/module.c
@@ -44,7 +44,6 @@ ecs_entity_t ecs_import(
 
     char *path = flecs_module_path_from_c(module_name);
     ecs_entity_t e = ecs_lookup(world, path);
-    ecs_os_free(path);
     
     if (!e) {
         ecs_trace("#[magenta]import#[reset] %s", module_name);
@@ -54,7 +53,7 @@ ecs_entity_t ecs_import(
         module(world);
 
         /* Lookup module entity (must be registered by module) */
-        e = ecs_lookup(world, module_name);
+        e = ecs_lookup(world, path);
         ecs_check(e != 0, ECS_MODULE_UNDEFINED, "%s", module_name);
 
         ecs_log_pop();
@@ -62,10 +61,12 @@ ecs_entity_t ecs_import(
 
     /* Restore to previous state */
     ecs_set_scope(world, old_scope);
+    ecs_os_free(path);
     world->info.name_prefix = old_name_prefix;
 
     return e;
 error:
+    ecs_os_free(path);
     return 0;
 }
 


SanderMertens:
Fixed with #2086! Ended up populating the dl callbacks in the OS API as well, so I could verify the behavior is correct with tests.

Issue #2094, opened 2026-05-14
Tick Source should be able to tick more than once per frame


jknightdoeswork:
EDIT:

I've implemented this in a [PR](https://github.com/SanderMertens/flecs/pull/2102) that adds a `ecs_set_fixed_interval()`

**Describe the problem you are trying to solve.**

I'm trying to use the flecs pipeline and system api to build a core loop that requires three distinct concepts to be expressed:

```
PreSimulation   (once per frame)
FixedSimulation (zero to N times per frame)
PostSimulation  (once per frame)
```

`FixedSimulation` runs core simulation logic at a fixed timestep (e.g. 1/60s) using a time accumulator. `PreSimulation` handles eg) input, and `PostSimulation` handles eg) rendering.

The key requirement is that the `FixedSimulation` must be able to tick **multiple times in a single frame** when the frame's delta time >= 2 * interval.

eg) Imagine a timer with interval 1.0 seconds, if you call ecs_progress with 10 seconds, the timer fires once with a delta time of 10, but I expect it to fire 10 times with a delta time of 1.0.

This is achievable today, but only by splitting the loop across three separate pipelines and implementing the fixed interval logic manually:

```c
ecs_run_pipeline(world, PreSimulation, raw_delta_time);

const float fixed_dt = 1.0f / 60.0f;
accumulator += raw_delta_time;
while (accumulator >= fixed_dt) {
    ecs_run_pipeline(world, FixedSimulation, fixed_dt);
    accumulator -= fixed_dt;
}

ecs_run_pipeline(world, PostSimulation, raw_delta_time);
```

This works, but it bypasses `ecs_progress` entirely and requires the user to manually manage what should be expressible within a single pipeline using tick sources.

**Describe the solution you'd like**

The current `tick_source` / rate filter API only supports ticking a phase 0 or 1 times per `ecs_progress` call. If the frame's delta time is large enough to warrant multiple fixed-timestep iterations, the tick source will still only fire once, causing the simulation to fall behind.

I'd like `tick_source` to support ticking a phase **0 to N times** per frame based on accumulated time, so that a single pipeline with phases like this would accurately simulate real time no matter what delta_time is passed into ecs_progress().

```
PreSimulation  (ticks once per frame)
FixedSimulation (fixed interval tick_source - may tick 0–N times)
PostSimulation (ticks once per frame)
```

This would eliminate the need for the three-pipeline workaround and let `ecs_progress` drive the entire core loop as intended.

[...]

PR #2102, opened 2026-05-16
Fixed Tick Rate Timer

jknightdoeswork:
Adds a new set of methods allowing a 'fixed interval' timer to be created and managed.

Core new method is
```
ecs_set_fixed_interval(world, entity_id, fixed_interval)
```

This PR adjusts tick sources to be able to tick multiple times per pipeline run with a fixed delta time.

Addresses this [issue](https://github.com/SanderMertens/flecs/issues/2094).

I think this should improve the ergonomics of defining a core game loop that requires parts of the pipeline to run at a fixed interval.

I think this test demonstrates the core difference:

```
void Timer_fixed_interval(void) {

    ecs_entity_t timer = ecs_set_fixed_interval(world, 0, 0.5);

    ecs_progress(world, 10.0);
    const EcsTickSource *src = ecs_get(world, timer, EcsTickSource);
    test_assert(src != NULL);
    test_int(src->ticks, 20.0); // with regular interval, this ticks once
    test_flt(src->time_elapsed, 0.5);   // with regular interval, this is 10.0

    ecs_fini(world);
}
```

Looking for discussion on this approach.

My understanding of the built in timer mechanism is that if you use it, you expose yourself to perhaps unexpected behavior if you progress a pipeline with a longer duration than 2.0*timer_interval. There is no built in mechanism from what I can tell in flecs that allows a system to be gaurenteed to be called with a fixed delta time. 

This PR additively allows this key concept in a simulation loop to be expressed elegantly using the built in flecs pipeline and timers APIs.

My instinct is that this should be the default mechanics of timers, but this is written to be backwards compatible, and does not change timers or systems at all unless the new fixed_interval set of APIs are used.

Let me know if I'm missing something or my understanding is incorrect!

Thanks for the awesome library!

### DIFF
--- a/src/addons/system/system.c
+++ b/src/addons/system/system.c
@@ -45,7 +45,7 @@ ecs_entity_t flecs_run_system(
             time_elapsed = tick->time_elapsed;
 
             /* If timer hasn't fired we shouldn't run the system */
-            if (!tick->tick) {
+            if (tick->ticks == 0) {
                 return 0;
             }
         } else {
--- a/src/addons/system/system.c
+++ b/src/addons/system/system.c
@@ -31,6 +31,7 @@ ecs_entity_t flecs_run_system(
 {
     flecs_poly_assert(world, ecs_world_t);
     ecs_ftime_t time_elapsed = delta_time;
+    uint32_t num_ticks = 1;
     ecs_entity_t tick_source = system_data->tick_source;
 
     /* Support legacy behavior */
@@ -38,17 +39,15 @@ ecs_entity_t flecs_run_system(
         param = system_data->ctx;
     }
 
-    uint32_t num_ticks = 1;
     if (tick_source) {
         const EcsTickSource *tick = ecs_get(world, tick_source, EcsTickSource);
 
         if (tick) {
-            time_elapsed = tick->time_elapsed;
-
             /* If timer hasn't fired we shouldn't run the system */
             if (tick->ticks == 0) {
                 return 0;
             }
+            time_elapsed = tick->time_elapsed;
             num_ticks = tick->ticks;
         } else {
             /* If a timer has been set but the timer entity does not have the
--- a/src/addons/system/system.c
+++ b/src/addons/system/system.c
@@ -83,6 +83,8 @@ ecs_entity_t flecs_run_system(
 
     flecs_poly_assert(stage, ecs_stage_t);
 
+    ecs_entity_t old_system = flecs_stage_set_system(stage, system);
+    ecs_entity_t interrupted_by = 0;
     for (uint32_t i = 0; i < num_ticks; i++) {
         /* Prepare the query iterator */
         ecs_iter_t wit, qit = ecs_query_iter(thread_ctx, system_data->query);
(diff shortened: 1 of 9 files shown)


src/addons/system/system.c
```diff
@@ -81,67 +82,71 @@ ecs_entity_t flecs_run_system(
...
-    qit.ctx = system_data->ctx;
-    qit.callback_ctx = system_data->callback_ctx;
-    qit.run_ctx = system_data->run_ctx;
-
-    if (system_data->group_id_set) {
-        ecs_iter_set_group(&qit, system_data->group_id);
-    }
+    ecs_entity_t old_system = flecs_stage_set_system(stage, system);
+    ecs_entity_t interrupted_by = 0;
+    for (uint32_t i = 0; i < num_ticks; i++) {
```
jknightdoeswork, line 87: The difference here is that the system is called in a loop, once per num_ticks.

src/addons/system/system.c
```diff
@@ -81,67 +82,71 @@ ecs_entity_t flecs_run_system(
...
-                    action(it);
-                }
+                action(&qit);
+                ecs_iter_fini(&qit);
             }
-        } else {
-            action(&qit);
-            ecs_iter_fini(&qit);
         }
+        interrupted_by = it->interrupted_by;
```
jknightdoeswork, line 149: The most recent interrupted_by is returned - not sure if this is precisely the intent of interrupted by.

src/addons/timer.c
```diff
@@ -18,33 +17,44 @@ void ProgressTimers(ecs_iter_t *it) {
...
-            }
-
-            timer[i].time = t; /* Initialize with remainder */
-            tick_source[i].tick = true;
-            tick_source[i].time_elapsed = time_elapsed - timer[i].overshoot;
-            timer[i].overshoot = t;
-
-            if (timer[i].single_shot) {
-                timer[i].active = false;
+        if (timer[i].fixed_interval) {
```
jknightdoeswork, line 26: if fixed_interval is false, the semantics are exactly the same as before.

src/addons/timer.c
```diff
@@ -140,6 +150,34 @@ ecs_ftime_t ecs_get_timeout(
...
+        .fixed_interval = true,
+        .active = true
+    });
+
+    ecs_system_t *system_data = flecs_poly_get(world, timer, ecs_system_t);
+    if (system_data) {
+        system_data->tick_source = timer;
+    }
+
+error:
```
jknightdoeswork, line 177: Full disclosure: I directly copy'd this `error` label pattern without precisely tracking where the error can be jumped from. I will review this personally more precisely before merging but requesting comment.

src/addons/timer.c
```diff
@@ -165,6 +203,32 @@ ecs_entity_t ecs_set_interval(
...
+    t->timeout = interval;
+    t->active = true;
+    t->fixed_interval = true;
+    ecs_modified(world, timer, EcsTimer);
+
+    ecs_system_t *system_data = flecs_poly_get(world, timer, ecs_system_t);
+    if (system_data) {
+        system_data->tick_source = timer;
+    }
+error:
```
jknightdoeswork, line 228: See above regarding not precisely understanding this `error` pattern.

PR #2113, opened 2026-05-27
Implement component inheritance as a storage feature

SanderMertens:
This PR makes component inheritance work for most flecs APIs, in addition to queries (which already works):

```cpp
world.component<Base>();
world.component<Foo>().is_a<Base>();
world.component<Bar>().is_a<Base>();

flecs::entity e = world.entity("e")
  .add<Foo>()
  .add<Bar>();

e.has<Base>(); // true
Base& b = e.get<Base>(); // returns first matching
```

This PR implements the following changes:
- sink component inheritance into the storage, so it becomes available for all APIs
- `get`, `has`, `owns`, `each` now work with inheritance
- queries for inherited components are now cacheable
- queries for inherited components now use the correct stride when matching a derived type
- inheritance support is added to reflection, so types can inherit from each other


src/query/engine/engine.h
```diff
@@ -114,6 +114,12 @@ int16_t flecs_query_next_column(
     ecs_id_t id,
     int32_t column);
 
+int16_t flecs_query_next_inherited_column(
```
SanderMertens: This should be integrated with ecs_search.

include/flecs/addons/cpp/delegate.hpp
```diff
@@ -188,6 +210,70 @@ struct each_field<T, if_t< is_pointer<T>::value &&
     }
 };
 
+template <typename T, typename = int>
+struct each_field_strided { };
+
+template <typename T>
```
SanderMertens, line 196: Investigate binary size increase.

src/query/engine/eval.c
```diff
@@ -17,6 +17,20 @@ bool flecs_query_dispatch(
...
 
+static
+int16_t flecs_query_next_column_w_inherited(
+    const ecs_world_t *world,
+    ecs_table_t *table,
+    const ecs_table_record_t *tr,
+    ecs_id_t id,
+    int32_t column)
+{
+    if (!ecs_id_match(table->type.array[tr->index], tr->hdr.cr->id)) {
```
SanderMertens: When flecs_query_next_inherited_column is integrated with ecs_search, this branch should go away.

src/storage/table.c
```diff
@@ -844,11 +846,282 @@ void flecs_table_init(
...
+        ecs_component_record_t *cr = tr->hdr.cr;
+        ecs_table_cache_remove(&cr->cache, table->id, &tr->hdr);
+        flecs_wfree_t(world, ecs_table_record_t, tr);
+        flecs_component_release(world, cr);
+    }
+
+    ecs_vec_clear(v);
+}
+
+static
```
SanderMertens: Move flecs_inheritance functions to a new component_inheritance.c file.

src/iter.c
```diff
@@ -125,8 +125,8 @@ void* ecs_field_w_size(
 
     int16_t column = it->columns[index];
     if (column >= 0) {
-        return ECS_ELEM(it->table->data.columns[column].data,
-            (ecs_size_t)size, it->offset);
+        ecs_column_t *col = &it->table->data.columns[column];
+        return ECS_ELEM(col->data, col->ti->size, it->offset);
```
SanderMertens, line 171: Revert this to using the passed in size. Assert that the size passed in to the function equals the column size. Add a separate ecs_base_field function that does use col->ti->size.

src/observable.c
```diff
@@ -1454,6 +1454,22 @@ void flecs_emit(
                      * any new reachable components. */
                     er_fwd = er_onset;
                 }
+
+                if (!table_event) {
```
SanderMertens: Should be OK since count is typically going to be 1, but should gate this on only propagating for entities that are queried for.

test/core/src/ComponentInheritance.c
```diff
@@ -0,0 +1,1643 @@
...
+    ECS_COMPONENT(world, Unit);
+    ECS_COMPONENT(world, Warrior);
+
+    ecs_entity_t e = ecs_new(world);
+    ecs_set(world, e, Warrior, {.hp = 10, .dmg = 5});
+
+    test_assert(ecs_get(world, e, Unit) == NULL);
+    test_bool(ecs_has_id(world, e, ecs_id(Unit)), false);
+
+    ecs_add_pair(world, ecs_id(Warrior), EcsIsA, ecs_id(Unit));
```
SanderMertens, line 383: This does not need to be supported, as it modifies an in-use component.

examples/cpp/queries/component_inheritance/src/main.cpp
```diff
@@ -1,56 +1,45 @@
...
+// inheritance trees built from tags instead of components.
+
+// Buf is the base component. HealthBuf and ManaBuf derive from it, so they
+// inherit the base members (the base subobject is stored at the start of the
+// derived component).
+struct Buf {
+    float value;
+};
+
+struct HealthBuf : Buf { };
```
SanderMertens, line 17: Create types that add members on top of the base type (same for C example).

src/query/engine/eval.c
```diff
@@ -48,24 +48,51 @@ bool flecs_query_select_w_id(
         }
     }
 
-repeat:
-    if (!redo || (op_ctx->remaining <= 0)) {
+repeat: {
+    bool next_table = !redo;
+    if (redo) {
```
SanderMertens, line 53: This code should not be necessary since flecs_query_next_column uses ecs_search, which supports inheritance.

src/bootstrap.c
```diff
@@ -242,6 +242,80 @@ void flecs_register_flag_for_trait(
     return;
 }
 
+static
```
SanderMertens, line 245: A global observer on IsA pairs is expensive. Move this to component record creation.

SanderMertens:
Landed in different PR

PR #2171, opened 2026-06-30
Fix assert on same-table name-index reparent in deferred batch

simon112:
(bug in upstream flecs found by AI subagent while hunting a different bug that seems not to be in upstream flecs. I do not understand what it's about but trust it to have likely found something real. AI remarks follow:) 

A deferred command batch that removes and re-adds the same exclusive pair — e.g. `remove(ChildOf, parent)` then `add(ChildOf, parent)` to move a named child to the end of its ordered siblings — ends the batch with the entity in its original table. `flecs_reparent_name_index` then computes `src_index == dst_index` and trips `ecs_assert(src_index != dst_index, ...)`, aborting in debug builds. Release builds were unaffected: the reparent is a benign remove-and-re-add of the same entity in the same name hashmap.

Replace the assert with a graceful early return when the indices are equal — the entity's name entry does not need to move. This subsumes the existing both-NULL early return.

Repro (debug build): a named child under a parent with `EcsOrderedChildren`, `ecs_remove_pair` + `ecs_add_pair` of `(ChildOf, parent)` inside one `ecs_defer_begin`/`ecs_defer_end`, aborts at `entity_name.c` without this fix.

Adds regression test
`Hierarchies_defer_batch_remove_add_childof_same_pair_named`.

(additional AI remark: "One judgment note: I kept the PR to just the debug-assert fix — the minimal, uncontroversial change. The deeper "should deferred remove+add of an identical exclusive pair be a no-op vs. move-to-end?" semantics question (which the subagent flagged) is deliberately not in this PR; that's a maintainer design call better raised as an issue if you want to pursue it.")

### DIFF
--- a/distr/flecs.c
+++ b/distr/flecs.c
@@ -11661,8 +11661,9 @@ void flecs_reparent_name_index(
 
     ecs_hashmap_t *src_index = flecs_table_get_name_index(world, src);
     ecs_hashmap_t *dst_index = flecs_table_get_name_index(world, dst);
-    ecs_assert(src_index != dst_index, ECS_INTERNAL_ERROR, NULL);
-    if ((!src_index && !dst_index)) {
+    if (src_index == dst_index) {
+        /* Both tables share a name index, so the entry doesn't need to move.
+         * Also covers the case where neither table has a name index. */
         return;
     }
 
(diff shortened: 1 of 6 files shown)


SanderMertens:
LGTM!

PR #2182, opened 2026-07-01
Fix OrderedChildren: track children added after enabling trait

simon112:
(bug discovery and all content of this PR other than in these parentheses from AI)

Adding EcsOrderedChildren to a parent that already has children left the existing child tables unflagged (EcsTableHasOrderedChildren), so any child added afterwards held its (ChildOf, parent) pair but was never appended to ordered_children — silently omitted from ecs_get_ordered_children / ecs_children. Adding the trait before any children exist worked, so the bug only affected the "enable ordering on an already-populated parent" path.

The EcsOnAdd hook flecs_register_ordered_children inited + populated the ordered vec and set the cr flag, but (unlike flecs_component_ordered_children_init) never iterated cr->cache to flag existing child tables. Route the hook through flecs_component_ordered_children_init so those tables are flagged and later child moves into them are appended.

Regression test: OrderedChildren.add_child_after_late_ordered_children.

### DIFF
--- a/distr/flecs.c
+++ b/distr/flecs.c
@@ -4700,9 +4700,10 @@ void flecs_register_ordered_children(ecs_iter_t *it) {
             ecs_component_record_t *cr = flecs_components_ensure(
                 it->world, ecs_childof(parent));
             if (!(cr->flags & EcsIdOrderedChildren)) {
-                flecs_ordered_children_init(it->world, cr);
+                /* Also flags existing child tables so children added after the
+                 * trait are tracked in the ordered list. */
+                flecs_component_ordered_children_init(it->world, cr);
                 flecs_ordered_children_populate(it->world, cr);
-                cr->flags |= EcsIdOrderedChildren;
             }
         }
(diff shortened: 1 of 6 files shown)


SanderMertens:
~~Did you verify this against the latest version of Flecs? I have existing tests for this.~~ Looks like this is failing on latest, taking a look!

SanderMertens:
One note: I'm not fundamentally against AI-generated PRs as long as they fix real issues and code quality is good. A problem with this PR (and the other one as well) is that it generates comments that are way too long, often too specific as its only focuses on the issue it's fixing, and context of the process it uses to fix the bug sometimes leak into comments.

If you want your AI-generated PRs to land, make sure that they are consistent with existing code, which means getting rid of these overly verbose AI generated comments.

simon112:
I had Claude rewrite the comments in view of your feedback. Let me know if there are any other issues that should be fixed.

SanderMertens:
LGTM!

PR #2230, opened 2026-07-21
[observers] Fix observers incorrectly evaluating value-pair targets as variables.

cryticx:


SanderMertens:
Can you add a test case to this PR that demonstrates what the code fixes?

cryticx:
Fixed by Sander separately from this PR, so closing as irrelevant.

PR #2241, opened 2026-08-01
fix: restore observer state after traversal mismatch

ShiroKSH:
## Summary

- Filter traversal-mismatched observers before changing stage or iterator state.
- Add a regression assertion for batched events propagated through multiple traversable relationships.
- Regenerate the distribution sources.

## Root cause

An expected traversal mismatch returned after setting the root stage's current observer and increasing log nesting. That early return skipped the existing cleanup path, leaving the root stage attributed to the skipped observer.

## Testing

- `bake run test/core -- Observer.2_up_terms_w_batched_add`
- `bake run test/core -- Observer`
- `clang -std=gnu99 -fsyntax-only -Werror -Wall -Wextra -Wno-unused-parameter -Wno-missing-field-initializers distr/flecs_no_addons.c`

### DIFF
--- a/distr/flecs.c
+++ b/distr/flecs.c
@@ -15646,6 +15646,15 @@ static void flecs_uni_observer_invoke(
         return;
     }
 
+    ecs_query_t *query = o->query;
+    if (query) {
+        ecs_term_t *term = &query->terms[0];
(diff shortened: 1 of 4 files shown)


SanderMertens:
The test relies on internal state. To prove this is a real bug, please provide a reproducer that only uses public APIs. Closing this for now.

PR #2242, opened 2026-08-01
fix: bound HTTP query parameters

ShiroKSH:
## Summary

- Limit parsed query parameters to `ECS_HTTP_QUERY_PARAM_COUNT_MAX`.
- Preserve the final accepted parameter when additional query fields are present.
- Add HTTP regression coverage and regenerate the distribution source.

## Root cause

The request parser recorded every query parameter although `ecs_http_request_t::params` is a fixed 32-entry array. Decoding a request with a 33rd field wrote past that array.

## Testing

- `bake run test/addons -- Http.query_params_max`
- `bake run test/addons --cfg sanitize -- Http.query_params_max`
- `bake run test/addons -- Http`

### DIFF
--- a/distr/flecs.c
+++ b/distr/flecs.c
@@ -46536,6 +46536,12 @@ static bool http_header_writable(
     return frag->header_count < ECS_HTTP_HEADER_COUNT_MAX;
 }
 
+static bool http_param_writable(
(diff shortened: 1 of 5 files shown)


SanderMertens:
LGTM!

PR #2243, opened 2026-08-01
fix: validate sparse id generations

ShiroKSH:
## Summary

- Verify sparse IDs against their dense entry before reporting, reading, or removing them.
- Return false for stale generations instead of removing the recycled entry.
- Add regression coverage for a recycled sparse ID and regenerate both distribution sources.

## Root cause

Sparse entries are indexed by the lower 32 bits of an ID. After an entry is recycled with a newer generation, the old ID could still pass liveness checks and read or remove the new entry because the dense ID was not compared with the requested full ID.

## Testing

- `bake run test/collections -- Sparse.generation_validation`
- `bake run test/collections -- Sparse`
- `clang -std=gnu99 -fsyntax-only -Werror -Wall -Wextra -Wno-unused-parameter -Wno-missing-field-initializers distr/flecs_no_addons.c`

### DIFF
--- a/distr/flecs.c
+++ b/distr/flecs.c
@@ -32930,6 +32930,11 @@ bool flecs_sparse_remove(
     int32_t dense = page->sparse[offset];
 
     if (dense) {
+        uint64_t *dense_array = ecs_vec_first_t(&sparse->dense, uint64_t);
+        if ((dense >= sparse->count) || (dense_array[dense] != id)) {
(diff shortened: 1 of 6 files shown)


SanderMertens:
The sparse data structure is only used for internals, and generation checks are in places where it is used. If you think there is a bug, please find a scenario that exercises the public APIs.

